PR #12, opened 2016-06-16
Metafields for Front Matter

mertkahyaoglu:
- [x] Add metafields that can be simple, array or object type
- [x] Convertion between types
- [x] Remove fields
- [x] Structure metafields when a meta object is given
- [x] Tests
- [x] Documentation
- [x] Editable `path` at the top
- [x] Show datepicker if field key is `date`
- [x] Sort array fields


mertkahyaoglu:
@benbalter I think I'm not authorized now 😁

parkr:
@mertkahyaoglu Not authorized to trigger builds on Travis? You should be. Can you visit the page and everything? Maybe try merging master if you haven't already? Hm.

mertkahyaoglu:
@parkr It says "You're not authorized to merge this pull request."

parkr:
@mertkahyaoglu That's very strange... Does it give a reason? Just not having enough permissions? It says you should have "Write" permissions to the repo according to our team permissions.

parkr:
@mertkahyaoglu Ah, from @benbalter (https://github.com/jekyll/jekyll-admin/pull/13#issuecomment-227295127):

> Also set up protected branches, with Travis being a required status check.

mertkahyaoglu:
Ah, I missed that. OK then 😀

mertkahyaoglu:
@benbalter This is ready! (except DOM testing, I will create another PR for other missing DOM tests as well.) Looking forward to your feedback.

I have one question. Currently, in edit view, we have `title` at the top, date and path at the sidebar and rest of the front matter at the bottom in metafields. Which attributes should be visible initially? For example, we don't need to show `layout` attribute in the metafields. When they are editing a page, it should be `page` by default. They can add an extra field if they would like to edit it. The same as `permalink`, as you said earlier.

benbalter:
* The title of the page/document is displayed twice, once editable, once not.

  screen_shot_2016-06-21_at_10_47_56_am
  
  I assume the 2nd one is the title via the YAML front matter? I'm thinking the first one should be the filename, with extension, also editable, e.g.

  screen shot 2016-06-21 at 10 51 13 am

  Related, does the API account for renaming a page/document? Do we need to document that?

* It's fine if it's not implemented yet, but path is always blank (even though, logically, the document has a path). Per above, I'd maybe move it to the top of the page (similar to editing a file on github.com).

* There should be no date field for pages, only documents.

* For posts, does changing the date update the filename, or only the YAML-frontmatter date?

* Why does the order of the YAML matter? I feel like the drag/drop just adds complexity for both you and the user.

* Nice job with the list/object rendering and adding. Really well done. 

* I'd just be consistent with the labeling. What's a simple? Is it a key/value pair? For top-level values, you call them objects (which I think makes sense), but for nested objects, they're called key/value pairs? Again, I'd just be consistent.

* The save button wasn't working for me (it's fine if it's not implemented yet)

> Which attributes should be visible initially?

See above. I think the path should be implicit as part of the UI and its own field. We may be able to remove the date field as well, and either just let the user user the YAML editor, or if the top level key is called `date` (added by the user), show the date picker for the value. I suspect the majority of pages will use the date in the filename, and it might be confusing to also have a date picker so prominent.

Definately out of scope for this pull request (it'd be great polish to add towards the end of the project), but `prose.io` had a concept of defining which fields you wanted visible via the `_config.yml` file. Here, a user might have something like this in their config, which we'd read in to automatically propagate YAML keys when you edit/create a post:

```yml
excerpt: simple
author: simple
categories: list
something_complicated: object
```

benbalter:
Also, if Travis is green, you should be able to merge. Please let me know if that's not the case.

mertkahyaoglu:
> I assume the 2nd one is the title via the YAML front matter?

Yes, it is. It's a good idea to make the first one editable like `github.com` so I'm removing `path` from sidebar.

> does the API account for renaming a page/document?

We don't need to. PUT endpoints should handle it.

> For posts, does changing the date update the filename, or only the YAML-frontmatter date?

It depends on the server. If server sends the updated post with updated filename because frontmatter date is changed/added as a response to PUT request, front-end will display the new filename.

> Why does the order of the YAML matter? 

Sorting objects does not update the state, only for UI since it doesn't matter where your keys at but for arrays, it does. Should we completely remove sorting or make it functional for only arrays?

> What's a simple?

Basically value of each key is simple. When you create an object, its value will be simple and also for arrays each item is simple initially. You can convert array item to object though. Then you can add new key/value pairs for that array item. For example;

``` javascript
{
  "layout": "post", // `layout` is field name, its value is `simple` which is post
  "categories": "gsoc",
  "students": [ // here `students` is field name, but its value is array 
    { name: "Ankur" }, // this array item is object (key/value pair)
    "Mert" // but this one is simple
  ]
}
```

> The save button wasn't working for me (it's fine if it's not implemented yet).

I will implement the update functionality when the components are completed.

> if the top level key is called date (added by the user), show the date picker for the value

This sounds better. I can remove `date` from the sidebar (this makes our `advanced options` section completely gone) and let the user define it in the metafields. If the key is `date` we show datepicker 👍 .

> concept of defining which fields you wanted visible via the _config.yml file

Yeah, we could do something like this 👍  

> Also, if Travis is green, you should be able to merge

Yes, I can merge now 😋

benbalter:
> Should we completely remove sorting or make it functional for only arrays?

I like the idea of making it functional only for arrays (where order can matter). Otherwise, it's just added cruft/chrome.

PR #15, opened 2016-06-22
Server changes

rush-skills:
- [x] Added some basic info to gemspec and README
- [x] Added admin route (`/admin`) for frontend files
- [x] Added a basic api servlet to handle GET and POST on `/api`
- [x] Configuration GET and POST
- [x] Pages INDEX, SHOW, UPDATE, DELETE
- [x] Fix rubocop violations
- [x] Collections INDEX, Documents INDEX, SHOW, POST, DELETE
- [x] Docs
- [x] Tests 
- [ ]  Travis integration testing


lib/jekyll/admin.rb
```diff
@@ -1,7 +1,25 @@
...
+require "jekyll/admin/configuration/servlet"
+require "jekyll/admin/pages/handler"
+require "jekyll/admin/pages/servlet"
+require_relative "commands/serve"
 module Jekyll
   module Admin
-    # Your code goes here...
+    # Path to the compiled frontend folder
+    def self.public_path
+      File.join(File.dirname(Dir.pwd), 'jekyll-admin/lib/jekyll/admin/public/src').to_s
```
mertkahyaoglu: Front-end files compiled into `jekyll-admin/lib/jekyll/admin/public/dist`.

rush-skills: How are we going to put them there? Are we going to manually compile them before merging to master? Also, @mertkahyaoglu, can you add instructions on how to compile the files locally. I was trying to do that but ended up with a blank page (the start `npm start` works fine)

mertkahyaoglu: I just updated my comment before you wrote 😁  `npm run build` is what you want.

parkr:
@rush-skills This is looking great!! ✨

benbalter:
@rush-skills nice work. This is looking great! I have a few minor style suggestions (e.g., whitespace), but I'll open up a pull request to add Rubocop with Jekyll's default code style in a moment to automate that.

rush-skills:
@benbalter I have merged with master, ran rubocop and fixed all offenses. I had to ignore a few of them which where present in function definations of servlets, where do_GET was considered bad as it wasn't following snake_case, but that's the way WEBrick defines it

lib/jekyll/admin.rb
```diff
@@ -1,7 +1,26 @@
 require "jekyll/admin/version"
-
+require "jekyll/admin/admin_servlet"
+require "jekyll/admin/api_servlet"
+require "jekyll/admin/api_handler"
+require "jekyll/admin/configuration/handler"
+require "jekyll/admin/configuration/servlet"
+require "jekyll/admin/pages/handler"
+require "jekyll/admin/pages/servlet"
```
benbalter: If there's a lot of sub-files (which is fine), it may make sense to use [autoload](http://www.rubyinside.com/ruby-techniques-revealed-autoload-1652.html), e.g.:

```ruby
 module Jekyll
    module Admin
      autoload :AdminServlet, "jekyll/admin/admin_servlet"
    end
end
```

That way classes are lazy loaded, which may speed up the initial load a bit.

lib/jekyll/admin.rb
```diff
@@ -1,7 +1,26 @@
...
+require_relative "commands/serve"
 module Jekyll
   module Admin
-    # Your code goes here...
+    # Path to the compiled frontend folder
+    def self.public_path
+      File.join(File.dirname(Dir.pwd), 'jekyll-admin/lib/jekyll/admin/public/dist').to_s
```
benbalter: Rather than using `Dir.pwd`, I'd use `__FILE__` so that the path is relative to this exact file, rather than where the file's being invoked from.

lib/jekyll/admin.rb
```diff
@@ -1,7 +1,26 @@
...
+    end
+
+    # Returns a hash of route and the respective handler
+    def self.api_servlet_bindings
```
benbalter: Are these always going to get `Jekyll.sites.first`? If so, maybe it'd make more sense moving that into `serve.rb` to DRY things up a bit?

lib/jekyll/admin/api_handler.rb
```diff
@@ -0,0 +1,30 @@
+module Jekyll
+  module Admin
+    class ApiHandler
+      def initialize(site)
+        @site = site
+      end
+
+      def read_file(path)
+        File.read(path)
+      end
```
benbalter: For `_config.yml`, you may need to access the raw file to get it without Jekyll's defaults (which there may be a Jekyll helper method for), but for the most part, for pages, data, etc. we should be pulling it from the site object, not reading it from disk and re-parsing it ourselves... let Jekyll do the heavy lifting for consistency.

lib/jekyll/admin/api_handler.rb
```diff
@@ -0,0 +1,30 @@
...
+
+      def write_file(path, data)
+        File.write(path, data)
+      end
+
+      def delete_file(path)
+        File.delete(path)
+      end
+
+      def parse_frontmatter(url)
```
benbalter: Same here as above. Whenever possible, we want to use Jekyll's representation of the file, rather than reading and parsing it ourselves.

lib/jekyll/admin/api_servlet.rb
```diff
@@ -0,0 +1,43 @@
...
+module Jekyll
+  module Admin
+    class ApiServlet < WEBrick::HTTPServlet::AbstractServlet
+      def initialize(server, site)
+        super(server)
+        @site = site
+      end
+
+      def do_GET(_request, response) # rubocop:disable Style/MethodName
+        hash = { :site => @site, :data => "Hello World" }
```
benbalter: What's going on here?

lib/jekyll/admin/configuration/handler.rb
```diff
@@ -0,0 +1,22 @@
+require 'yaml'
+module Jekyll
+  module Admin
+    module Configuration
```
benbalter: Take a look at https://github.com/jekyll/jekyll/blob/master/lib/jekyll/configuration.rb. I think we'll want the raw config file, but Jekyll has some helpers to help you find its path and safely read it across different formats.

lib/jekyll/admin/pages/handler.rb
```diff
@@ -0,0 +1,46 @@
...
+    module Pages
+      class Handler < ApiHandler
+        def index
+          pages = []
+          @site.pages.each do |page|
+            next if page.data["title"].nil?
+            extname = File.extname(page.name)
+            page_id = File.basename(page.name, extname)
+            meta = parse_frontmatter(page.path)
+            page_item = {
```
benbalter: Whenever possible, I'd call `page.to_liquid.to_json` (to use Jekyll's native representation), even if we're then only taking a subset of that for e.g., indexes. That way our internal data models will be future proof to Jekyll's upstream changes.

lib/jekyll/admin/pages/handler.rb
```diff
@@ -0,0 +1,46 @@
...
+            }
+            pages.push page_item
+          end
+          pages
+        end
+
+        def show(file)
+          page = Jekyll::Page.new(@site, @site.source, "", file)
+          meta = parse_frontmatter(page.path)
+          {
```
benbalter: Same here. This should just be `page.to_liquid.to_json`.

benbalter:
@rush-skills this is starting to look good. I made a handful of line level comments.

One things that's obviously missing is documentation. While method-level documentation would be great, to start, how can I test this locally?

In terms of tests, I know there was some talk about integration tests, which is great, and something we should do in the long run, but in the interim, I believe you should be able to create unit-level Ruby tests for most of the methods by creating mock request and response objects, passing them to the method, and then asserting the response's output. Please let me know if you've tried that and that's not the case.

rush-skills:
@benbalter: the reason I am parsing the frontmatter manually is because I didn't find any method that directly gives me the hash of the metafields without adding the defaults. If we use #to_liquid/#data it adds defaults to the output which is not really the content of the file and thus the front-end representation of the file would differ from the actual one.
@parkr, @benbalter: I have added code for collections and method level documentation. I am working on unit followed by integrations tests and I am trying to figure out a way to setup a test project with this gem as a plugin inside the tests directory that I can run a server and/or tests on. Can you help he understand how should I script the test such that the files aren't actually changed, (as most post/delete calls actually read/write to the file system) and the same directory can be reused for all tests?

lib/jekyll/admin.rb
rush-skills: @benbalter autoload looks like a nice solution, but in my case most of the classes have the same name, i.e. Servlet and Handler. I am able to think of a way to make that work, can you help me out?

lib/jekyll/admin.rb
rush-skills: Fixed. Thanks, I now found that the previous version would have broken when this was installed as a gem.

lib/jekyll/admin.rb
rush-skills: Sorry, I didn' really get that. I have updated the code to use same site variable for all bindings in the same function, instead of calling `Jekyll.sites.first`. Why do you suggest moving it to serve.rb is a better solution? (lazy loading them?)

lib/jekyll/admin/api_servlet.rb
rush-skills: This was just a dummy servlet to serve up the `/api` route while I was figuring out how to make GETs and POSTs work in WEBrick Servlets. Now it serves the purpose of the default api page (on `/api` as other endpoints are on `/api/*` and I have updated it to show some help text instead.

lib/jekyll/admin/configuration/handler.rb
rush-skills: Thanks, I have updated the code to use those functions.

lib/jekyll/admin/collections/handler.rb
```diff
@@ -0,0 +1,135 @@
...
+        # meta - hash of metafields present in the document as frontmatter
+        # body - contents of the document, except frontmatter (in html)
+        def show(collection_name, document_id)
+          collection = @collections[collection_name]
+          doc = find_document(collection, document_id)
+          if doc
+            meta = parse_frontmatter(doc.path)
+            document = {
+              :collection_name => collection_name,
+              :document_id => doc.basename,
```
mertkahyaoglu: @parkr What is our final decision on these identifiers ? Should we make them only `id` ? Same as `page_id`.

parkr: @mertkahyaoglu Just `id` Is best in my opinion :smile:

rush-skills: Updated both methods to use `id` 😄

rush-skills:
@benbalter @parkr I have skipped unit tests and directly added integration tests using `airborne`. I setup a test folder in `test/test-site` and a script at `script/server` that clones that site to `test/source` and runs a jekyll server with `jekyll-admin` mounted, so that it can be easily developed and tested (till now i was using a project in different directory). I then setup integrations tests at `spec/*_spec.rb` which test the `/api` routes for the server running at `localhost:4000`. I have added development and testing instructions at README.

I have two problems with the current setup that I am trying to figure out a solution to:

1. Since the server needs to be started and stopped and manually and is supposed to be running at a fixed url, we can't run the tests on travis simply(can we?). 
2. The server doesn't reload on changes in the gem directory, has to be restarted manually.

mertkahyaoglu:
I have a question about our API. I think we can discuss it here. 
How does the server handle filename change? Since our `id`s are dependent on filenames, updating them should require a special care. One other thing is that how does the front-end know the filename of a document/page? (ids don't have extensions).

rush-skills:
@mertkahyaoglu As of this PR, I updated the IDs to have extensions (otherwise about.md and about.markdown being two different files would have the same id). To edit the filename, currently there is no method present (perhaps it should be an optional parameter in POST, something like new_id), so a hacky way would be to send DELETE on the old id and CREATE on the new one.

mertkahyaoglu:
@rush-skills Actually we can compare the `id`s that are one from the endpoint and the other from the request object. If they differ, we update the `id` with the requested one.

With allowing filename change, there comes another problem to solve. Users may try to create a document/page with the existing filename. That means we need another API endpoint to check for the filename/id.

mertkahyaoglu:
Can we close this one?

PR #19, opened 2016-06-27
Complete Content Editing

mertkahyaoglu:
This PR includes following features;

- [x] Contents can be updated now.
- [x] Save button is enabled when a form input changes and its text changes to `Saved` when updated.
- [x] Shows error messages if the form is not validated. (Only validation is the title being required for now. Other validations can be added.)
- [x] Replaced `page_id` and `document_id` with `id`
- [x] Renamings
- [x] Tests
- [x] Docs


mertkahyaoglu:
@benbalter Would love to know your thoughts on this 👀 

I have two questions regarding this PR.

1. I moved the `path` (filename) field to the top. It is editable now. My question is does updating filename/path update the `id`? I think it does. What should be the behavior after the filename/path is updated? We redirect user to the listing page? It is not user-friendly though. How can we solve this?

2. Currently, only the title field is required. If the user does not provide the title, an error message appears at the top. Which other fields should be validated? (body, filename..)

parkr:
❤️ ❤️ ❤️ ❤️ ❤️ 

> Currently, only the title field is required. If the user does not provide the title, an error message appears at the top. Which other fields should be validated? (body, filename..)

Personally, I think the `title` and `date` are the only two elements a post really _needs_, but you can default the `date` to today! I'm 👍 on just erroring for empty `title`. 😄

mertkahyaoglu:
@parkr Thanks for your feedback. You mean default the `date` to today when creating a post, right ? Because when updating, if `date` is present in the front matter, UI show it in the `metafields` section. Otherwise it does not. But when creating a post, I can put the date field showing the current date-time by default to the `metafields` section.

parkr:
> You mean default the date to today when creating a post, right ?

Yes! 😄 When editing a post, it should error if the `date` doesn't exist.

mertkahyaoglu:
Ah, got it. Users should not able to delete an existing date field as well when editing 👍

benbalter:
@mertkahyaoglu this is looking great. 

To clarify, `title` and `date` are required, but they're not required in the YAML front matter. If I have a post `_posts/2016-01-01-some-post.md`, Jekyll can pull the date and title from the file name. 

I can see an argument for always requiring the title field in the YAML front matter from a usability perspective, and am not strongly opinionated about requiring it or not, but I would caution against requiring the date in the YAML front matter as I think it sets a confusing default where the date in the filename could get out of sync with the date in the YAML front matter. For most users that care about dates of posts, and not specific times, the date in the filename should be sufficient for their needs.

mertkahyaoglu:
@benbalter So, for the front-end, it is enough to check for only title now, right? Another issue is filename is editable. Should we check for that as well whether it is in the format of yyyy-mm-dd-title.md for posts?

benbalter:
> Should we check for that as well whether it is in the format of yyyy-mm-dd-title.md for posts?

We can, as some nice polish, but I don't think it's necessary for the first pass at implementing the feature. It's a "nice to have" in my mind, not a "must have".

mertkahyaoglu:
@benbalter I think this regex might work `/^(.+\/)*(\d+-\d+-\d+)-(.*)(\.[^.]+)$/`. I don't know this is the one Jekyll uses. 

One other question is how do we edit filename? I know you did ask this in the previous PR but I thought changing `path` in the front matter would be enough. I guess we might need to edit the API.

parkr:
> @benbalter I think this regex might work /^(.+\/)*(\d+-\d+-\d+)-(.*)(\.[^.]+)$/. I don't know this is the one Jekyll uses.

@mertkahyaoglu We use [`Document::DATE_FILENAME_MATCHER`](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/document.rb#L12) 😄

mertkahyaoglu:
Oh, thanks @parkr 😁

mertkahyaoglu:
@parkr Shouldn't be a `\` after first plus? `!^(?:.+/)*(\d+-\d+-\d+)-(.*)(\.[^.]+)$`
I'm not that good with regex, teach me please 😁

parkr:
We're using the regexp literal style `%r` which doesn't use `/` as a delimiter so we don't need o escape `/` characters with `\/` like you would if it were `/<regexp>/`.

PR #28, opened 2016-07-06
WIP: Server fixes

benbalter:
This PR is intended to encompass a few, unrelated bug fixes to the API server that I noticed since it's been in use (feel free to comment with others if you know of any):

- [x] Always used `Jekyll.sanitized_path` to determine paths to files on disk (ensuring they're always in the site source)
- [x] Share `document_body` between documents and pages (YAML front matter + content)
- [x] Check that data files and static files exist in a way that parallels documents and pages
- [x] Have the configuration endpoint return the config from disk, rather than the computed config with defaults merged in
- [x] Call `Site#process` after any write operations so that previews remain accurate
- [ ] ~~Documents and Pages should return the YAML front matter from disk without defaults merged in~~


benbalter:
@parkr any creative ways to call `to_liquid` on both Documents and Pages without getting front matter defaults? The code doesn't make it look easy. One solution, although not ideal, would be to *remove* any front matter defaults from the result of the `to_liquid` call. It wouldn't be 100% accurate, but it'd be functionally the same.

parkr:
@benbalter Ha! This is totally not what `to_liquid` is meant for. Liquid being output-focused, we didn't design it for work with the source 😉  I believe the defaults are merged in in `Document#read`, actually, so you'd have to refactor that method... 

Actually, having defaults in the view would be nice. They could be greyed out the way default CSS values are in the Chrome Inspector, where if you click them, you can override the value. Users seeing a default `layout: post` when that value is set would be helpful when editing, I think.

benbalter:
> Actually, having defaults in the view would be nice. 

This problem, as I see it, is that editing a page or document wouldn't be independent.

1. Set a front matter default of `foo: bar` for posts
2. Open a post in the editor
3. Hit save, making no changes
4. File on disk now has `foo: bar` hard coded into the post

Perhaps the opposite (removing front matter defaults, if present) would be a good compromise. It'd be easier technically, and would have the added affect of removing duplicative front-matter.

For now, I'm going to punt on front matter defaults, to get the rest in, and will open a dedicated issue to track.

Issue #29, opened 2016-07-06
Allow CORS for dev server


mertkahyaoglu:
Frontend dev server runs on port 3000 and sends requests to the API running on port 4000. This is only for development. For production, both runs on the same port anyway. This check differentiates the API url accordingly;
``` javascript
if (process.env.NODE_ENV === 'production') {
  API = '/_api';
} else {
  API = 'http://localhost:4000/_api';
}
```
However, difference between ports cause CORS now. What can we do about this?

PR #31, opened 2016-07-07
Set CORS headers in development

benbalter:
@mertkahyaoglu this should set [these cross origin headers](https://github.com/britg/sinatra-cross_origin/blob/master/lib/sinatra/cross_origin.rb#L51-L57) in development, as well as properly respond to `OPTIONS` pre-flight checks. Does that look like it should work for testing between the two servers?

By default, JekyllAdmin will be in `production` mode (to prevent some strange cross-origin click jacking when running on a real site), but can be enabled by setting `RACK_ENV=development` (enabled by default for `script/test-server`).

Fixes https://github.com/jekyll/jekyll-admin/issues/29.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,12 @@ Upon successful installation, you should be able to access the admin panel at `/
 
 `script/test-server`
 
+### The environment flag
+
+When developing locally, it can be helpful to see error backtraces, disable template caching, have expanded request logs, and to allow cross-origin requests between the Ruby server and the Node server. By default, however, JekyllAdmin runs in `production` mode, meaning these development features are disabled.
+
+To enabled the development features, set the environmental variable `RACK_ENV` to `development`. When enabled, the `/_api/` endpoint will add `Access-Control-Allow-Origin: any` headers, and respond to `OPTIONS` pre-flight checks. This flag is set automatically when using the `script/test-server` command.
+
 ## Contributing
 
 Bug reports and pull requests are welcome on GitHub at https://github.com/jekyll/jekyll-admin. This project is intended to be a safe, welcoming space for collaboration, and contributors are expected to adhere to the [Contributor Covenant](http://contributor-covenant.org) code of conduct.
--- a/jekyll-admin.gemspec
+++ b/jekyll-admin.gemspec
@@ -35,4 +35,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "rake", "~> 10.0"
   spec.add_development_dependency "rspec", "~> 3.4"
   spec.add_development_dependency "rubocop", "~> 0.35"
+  spec.add_development_dependency "sinatra-cross_origin", "~> 0.3"
 end
--- a/lib/jekyll/admin.rb
+++ b/lib/jekyll/admin.rb
@@ -1,9 +1,9 @@
-require 'webrick'
-require 'sinatra'
-require 'sinatra/base'
+require "webrick"
+require "sinatra"
+require "sinatra/base"
 require "sinatra/json"
 require "sinatra/reloader"
-require 'json'
+require "json"
 require "jekyll"
 require "jekyll/admin/version"
 require "jekyll/admin/server"
--- a/lib/jekyll/admin/server.rb
+++ b/lib/jekyll/admin/server.rb
@@ -13,6 +13,14 @@ module Jekyll
         disable :allow_credentials
       end
 
+      ACCESS_CONTROL_ALLOW_HEADERS = %w(
+        X-Requested-With
+        X-HTTP-Method-Override
+        Content-Type
+        Cache-Control
+        Accept
+      ).freeze
+
       get "/" do
         json ROUTES.map { |route| ["#{route}_api", URI.join(base_url, "/_api/", route)] }.to_h
       end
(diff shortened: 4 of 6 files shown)


benbalter:
To note, one "gotcha" I ran into when testing locally, your request must set the `Origin` header, as specified by the CORS spec (which most javascript clients/browsers will do, but cURL won't).

parkr:
LGTM.

mertkahyaoglu:
@benbalter You're right. I can see the response properly from the browser but the `fetch` library is not able to get it. I tried setting `Origin`, still no work. 

Thank you, Ben. This PR works great. One problem I encountered is that I got a dependency error saying `'cannot load such file -- sinatra/cross_origin'`. 
I changed this line in `jekyll-admin.gemspec` file;

`spec.add_development_dependency "sinatra-cross_origin", "~> 0.3"` to this

`spec.add_dependency "sinatra-cross_origin", "~> 0.3"` and it works without any error.

benbalter:
> One problem I encountered is that I got a dependency error saying `'cannot load such file -- sinatra/cross_origin'`. 

That's going to be my fault. Fixed via https://github.com/jekyll/jekyll-admin/pull/32.

> I can see the response properly from the browser but the `fetch` library is not able to get it. I tried setting `Origin`, still no work. 

Can you explain what you're doing a bit more? Here's what I get via cURL:

```
$ curl -i -H "Origin: http://localhost:5000" http://localhost:4000/_api
HTTP/1.1 200 OK
Content-Type: application/json
Access-Control-Allow-Origin: http://localhost:5000
Access-Control-Allow-Methods: POST, GET, OPTIONS
Access-Control-Allow-Headers: *, Content-Type, Accept, AUTHORIZATION, Cache-Control
Access-Control-Allow-Credentials: false
Access-Control-Max-Age: 1728000
Access-Control-Expose-Headers: Cache-Control, Content-Language, Content-Type, Expires, Last-Modified, Pragma
Content-Length: 276
X-Content-Type-Options: nosniff
Server: WEBrick/1.3.1 (Ruby/2.3.1/2016-04-26)
Date: Thu, 07 Jul 2016 20:21:03 GMT
Connection: Keep-Alive
```

mertkahyaoglu:
Frontend is using `isomorphic-fetch` library for http calls. Here is how I was using it;

``` javascript
fetch(url, {
  method: 'GET',
  headers: {
    "Access-Control-Allow-Origin": "http://localhost:3000",
    "Access-Control-Allow-Methods": "GET"
  }
})
.then(res => res.json())
```

And the error message is;
```
Fetch API cannot load http://localhost:4000/_api/configuration. 
Response to preflight request doesn't pass access control check: 
No 'Access-Control-Allow-Origin' header is present on the requested resource. 
Origin 'http://localhost:3000' is therefore not allowed access.
```

benbalter:
/cc @jldec who may know more.

Could it possibly be related to https://github.com/matthew-andrews/isomorphic-fetch/issues/50?

It should definitely be `Origin` without the `Access-Control-Allow` on the request.

I don't know Node well. Could you provide a file I could run locally to recreate this?

We can always work around this, at least temporarily, by spoofing the request header on the server, but would obviously prefer to "do it right" if we can.

benbalter:
Alternatively, if you add `logger.info request.env` immediately after a `get "/WHATEVER"` line in ruby, run `script/test-server` and call that endpoint, the command line output should show the full request, including headers. You should see, e.g., `"HTTP_ORIGIN"=>"http://localhost:5000"` (headers are converted to `HTTP_*`).

benbalter:
Also, just to confirm, you're either running `script/test-server` or manually setting the `RACK_ENV=development` environmental flag based off of `master`?

mertkahyaoglu:
@benbalter I don't know If I misunderstood but after #31, it just works properly without setting headers. As I tried to mean it in my comment here 😁 

> Thank you, Ben. This PR works great.

Sorry If I couldn't express that but do you want me to reproduce the issue before merging #31? Now I have to set the flag to `production` in order to reproduce the same error.

mertkahyaoglu:
@benbalter I tried updating configuration. Server responded to the preflight request without allowing PUT method. I got `Method PUT is not allowed by Access-Control-Allow-Methods in preflight response.` error.

Here is the headers
screen shot 2016-07-08 at 11 24 13 pm

benbalter:
@mertkahyaoglu You are correct. I *think* I fixed it via https://github.com/jekyll/jekyll-admin/pull/33. Sorry for the trouble. Here's the response now:

```
$ curl -i -X OPTIONS -H "Origin: http://localhost:3000" http://localhost:4000/_api
HTTP/1.1 204 No Content
Access-Control-Allow-Origin: http://localhost:3000
Access-Control-Allow-Methods: DELETE, GET, OPTIONS, POST, PUT
Access-Control-Allow-Headers: *, Content-Type, Accept, AUTHORIZATION, Cache-Control
Access-Control-Allow-Credentials: false
Access-Control-Max-Age: 1728000
Access-Control-Expose-Headers: Cache-Control, Content-Language, Content-Type, Expires, Last-Modified, Pragma
X-Content-Type-Options: nosniff
Server: WEBrick/1.3.1 (Ruby/2.3.1/2016-04-26)
Date: Fri, 08 Jul 2016 21:15:00 GMT
Connection: Keep-Alive
```

mertkahyaoglu:
@benbalter Now it updates properly but I'm getting this error now.

`Fetch API cannot load http://localhost:4000/_api/configuration. The request was redirected to 'http://localhost:4000/_api/configuration', which is disallowed for cross-origin requests that require preflight.`

benbalter:
@mertkahyaoglu Aren't computers great? :smile: It sounds like this is on a `PUT`/`POST` request and that it's writing to disk, but erring before the response is processed? I believe this is due to the server issuing a `30x` redirect after the write, which CORS doesn't like. If that's the case, I'd concentrate on the read operations, and I will fix things on the server — I don't believe it's your fault.

PR #36, opened 2016-07-11
Don't redirect CORS requests after write operations

benbalter:
Per https://github.com/jekyll/jekyll-admin/pull/31#issuecomment-231491593, CORS requests can't result in a `30x` redirect.

This PR updates the logic such that any `PUT` or `POST` operation returns the resulting object, rather than a 301 to the corresponding `GET` endpoint.

Likewise, under the same rationale, the `GET` operation for static files, now returns the static file object, rather than a blind 301 to the real path (but the object contains the real path, so behavior can be preserved).


parkr:
Nicely done. ✨

Issue #41, opened 2016-07-11
Convert document IDs from `/YYYY/MM/DD/ID` to `YYYY-MM-DD-ID.md`


benbalter:
Per https://github.com/jekyll/jekyll-admin/issues/35#issuecomment-231766161.

parkr:
Is the API just using `document.id`? I'd let Jekyll handle this instead of normalizing.

Issue #30, opened 2016-07-07
Documents and Pages metadata contain frontmatter defaults


benbalter:
See https://github.com/jekyll/jekyll-admin/pull/28#issuecomment-231075285.

I'd like to propose that we *remove* front matter defaults, even if explicitly present in the front matter. It's a compromise that isn't too difficult to implement technically, is better than the alternative, and has the added affect of removing duplicative front matter data.

parkr:
> has the added affect of removing duplicative front matter data.

How so? Defaults are defaults – if you already have a `layout`, for example, then the default value won't be shown. It'd be useful to see all the metadata that the document has, including defaults. I think it shouldn't be editable, but it should be shown.

benbalter:
>  I think it shouldn't be editable, but it should be shown.

So we should parse the YAML from the document directly? How do we cleanly get that, and the other `to_liquid`-injected values (e.g., date), without getting the defaults?

Issue #24, opened 2016-07-05
Frontend should properly namespace API calls


benbalter:
Right now the front end server [assumes all API calls live in a top-level namespace](https://github.com/jekyll/jekyll-admin/blob/master/lib/jekyll/admin/public/src/constants/api.js#L6-L21), (e.g., `/collections).

The site may have an existing `/posts/` and `/pages` path that we don't want to clobber. To be safe, we should namespace the API endpoints. 

I'm suggesting `/_api/` (e.g., `/_api/posts`), since the `_` convention is an existing Jekyll convention, and users are unlikely to have a generated `_api` path.

PR #42, opened 2016-07-11
Adapt Frontend to the API

### LINKED ISSUE #23: Frontend can't assume assume port
Right now, the front end [assumes the server is running on `localhost:3004`](https://github.com/jekyll/jekyll-admin/blob/master/lib/jekyll/admin/public/src/constants/api.js#L6-L21). 

While the default server port will be `4000`, users can set that via a flag, so we shouldn't assume that either. 

Ideally, the server would get the port from the URL it's being requested from, or from the server directly.

mertkahyaoglu:
Closes #23, #24 

- [x] Update API endpoints
- [x] Remove mock API to use the real one
- [x] Adapt actions, reducers and components to the API
- [x] Fix tests
- [x] Replaces `POST` with `PUT` again
- [x] Fix list, edit and create views (depends on #30, #41)


mertkahyaoglu:
This PR depends on #30. I'm in favor of removing defaults as well because right now updating a document/page adds defaults to the front matter.

benbalter:
> Replaces POST with PUT again

If I did this backwards, I don't recall where things landed, and easy enough to change back on the server side, if you think it should be one way or the other.

mertkahyaoglu:
I also think PUT is better but had to change it cause of Ankur implementation.

benbalter:
> I also think PUT is better

If there's a change you'd like on the server side, I'd implement things the way you think it should be here, and open an issue describing the desired behavior, and we'll take care of it (I'm not sure which endpoints your referring to specifically).

mertkahyaoglu:
@benbalter Updating config adds `posts` to the `collections` section in the `_config.yml` file. Here are the steps to reproduce the issue.

1._config.yml includes the following content;

``` yaml
title: Your awesome title
collections:
  test:
    foo: bar
    output: true
```

2.Update the config

3._config.yml file now includes the following content;

``` yaml
title: Your awesome title
collections:
  posts:
    output: true
  test:
    foo: bar
    output: true
```

4.Delete the following lines;

``` yaml
posts:
  output: true
```

5.It deletes those lines from the file but the response to the front-end still has them.

benbalter:
@mertkahyaoglu would you mind opening a dedicated issue for what you just described (copy and paste is fine). That way we can track fixing that server-side bug regardless of the client-side implementation you're working on here (and then it shouldn't be seen as a blocker).

benbalter:
Chatted with @mertkahyaoglu via Slack, and although there's a few things still not yet implemented, this LGTM, so that we can keep subsequent PR's smaller.

PR #43, opened 2016-07-13
Return pages and documents without frontmatter defaults

benbalter:
Fixes https://github.com/jekyll/jekyll-admin/issues/30.

@parkr (or maybe @pathawks) before I go down the path of adding tests, I could use your 💭...

We want to be able to return pages and documents without their front matter defaults.

For Pages, that's a matter of making a new `to_liquid` method, that doesn't merge in front matter defaults, as the defaults are not merged in until `to_liquid` is called.

For Documents, that's a matter of re-reading the file, duplicating the object, and manually swapping in the data before calling `to_liquid`, because front matter defaults are injected on load.

Specific questions:

1. Is this the right approach? Is there a better one?
2. is their a ruby file naming convention for monkey patching?

### DIFF
--- a/lib/jekyll/admin.rb
+++ b/lib/jekyll/admin.rb
@@ -16,6 +16,7 @@ require "jekyll/admin/server/data.rb"
 require "jekyll/admin/server/page.rb"
 require "jekyll/admin/server/static_file.rb"
 require_relative "./commands/serve"
+require_relative "./convertible_ext"
 
 module Jekyll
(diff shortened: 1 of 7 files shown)


benbalter:
Going to go ahead and merge this to unblock https://github.com/jekyll/jekyll-admin/pull/42 for the time being, but don't know that this is the right approach. @parkr could use your 💭 still when you can.

PR #47, opened 2016-07-14
Support directory listing for /_api/static_files

### LINKED ISSUE #39: When `GET /static_files/:path` is called on a directory, it should return that directory's static files
Note: that this should be the directory's *static files*, as seen by Jekyll, not all files in the directory.

tekknolagi:
Fixes #39.


lib/jekyll/admin/server/static_file.rb
```diff
@@ -27,6 +30,10 @@ class Server < Sinatra::Base
 
         private
 
+        def set_file_path
+          params["static_file_id"] = params["splat"].first
```
benbalter: Could this be done in a `before '/*' do` block? I'm not sure if the splats are processed that early in the request? If not, perhaps we should just update the `static_file_path` logic to look at the splat? I don't *think* `static_file_id` is used anywhere else (and everything in this namespace is private to this namespace).

lib/jekyll/admin/server/static_file.rb
```diff
@@ -40,7 +47,16 @@ def static_files
         end
 
         def static_file
-          static_files.find { |f| f.path == static_file_path }
+          file = static_files.find { |f| f.path == static_file_path }
+          if not file
+            static_files.select do |f|
+              # Files that are in this directory
+              # Joined with / to ensure user can't do partial paths
+              f.path.start_with? File.join(static_file_path, "/")
```
benbalter: How would this handle matched folders? What about matched files in subfolders?

benbalter:
Nice start. We'll want to add tests, and I think there might be some ambiguity around how to handle sub-folders, that we'll have some leeway in deciding how to handle.

lib/jekyll/admin/server/static_file.rb
tekknolagi: It currently ignores folders as an entity and instead presents a full recursive tree listing of files from the folder specified on down.

lib/jekyll/admin/server/static_file.rb
tekknolagi: Yep, can look into this.

lib/jekyll/admin/server/static_file.rb
```diff
@@ -40,7 +43,21 @@ def static_files
         end
 
         def static_file
-          static_files.find { |f| f.path == static_file_path }
+          file = static_files.find { |f| f.path == static_file_path }
+          if !file
+            file_list = static_files.select do |f|
```
benbalter: I'd break `file_list` into it's own method. `static_file`'s getting a bit hard to read.

lib/jekyll/admin/server/static_file.rb
```diff
@@ -40,7 +43,21 @@ def static_files
...
+              # Files that are in this directory
+              # Joined with / to ensure user can't do partial paths
+              f.path.start_with? File.join(static_file_path, "/")
+            end.map(&:to_liquid)
+            if file_list == []
+              nil
```
benbalter: Should this always return an array? (even if empty)

benbalter:
@tekknolagi this is looking great. Had two minor style suggestions, and then think this should be good to go.

benbalter:
@tekknolagi any interest in implementing the feedback above so we can get this merged?

lib/jekyll/admin/server/static_file.rb
tekknolagi: @benbalter I think I added that because `find` returns `nil` if no results are found.

lib/jekyll/admin/server/static_file.rb
tekknolagi: Done.

tekknolagi:
@benbalter still have no tests written and have not figured out how to handle subfolder ambiguity. Thoughts?

benbalter:
@tekknolagi it looks like you've implemented tests (and they look great!). Can you explain what you mean about subfolder ambiguity? Thinking it may make sense to merge this as is, and then can polish in a subsequent, smaller pass.

tekknolagi:
@benbalter Whoa, forgot about those. I meant that I think listing `/folder` will list every static descendant of `/folder`, not just the immediate ones.

tekknolagi:
I'm fine with that.

benbalter:
Nice work. LGTM.

benbalter:
> will list every static descendant of /folder, not just the immediate ones. I'm fine with merging, though.

Better to have imperfect code that gets us 90% of the way there, than no code that gets us 0% of the way, IMHO. @mertkahyaoglu this should implement static file listings for you, and fixes issues retrieving static files in subfolders.

tekknolagi:
Fair point.

mertkahyaoglu:
👍  🎉

Issue #35, opened 2016-07-09
Document object should include unrendered markdown content


mertkahyaoglu:
I can fetch the all post documents. This is the result; (I added the `title` attribute to the front matter)

![screen shot 2016-07-09 at 5 43 41 pm](https://cloud.githubusercontent.com/assets/7414026/16708450/03f653a0-45fd-11e6-9fdc-b5003dbda494.png)

As you see, the response includes the document body (in html format but we have a markdown editor in the front-end).
Did we eliminate the `GET /collections/:collection_name/documents/:id`? Because I'm not able to fetch the post with id `/2016/01/01/test`. Although it is pointless since we have everything we need, I think.

benbalter:
@parkr any clean way to get the pre-rendered HTML with `to_liquid`? Shouldn't that be the difference between `output` and `content`?

> Did we eliminate the `GET /collections/:collection_name/documents/:id`? Because I'm not able to fetch the post with id `/2016/01/01/test`

I ran `script/test-server` (pro-tip, at least in Chrome, you can go to `/_api` and click through the resources if you have [this extension](https://chrome.google.com/webstore/detail/json-formatter/bcjindcccaagfpapjjmafapmmgkkhgoa?hl=en) installed), and was able to access `GET http://localhost:4000/_api/collections/posts/2016-01-01-test.md`.

I think `/collections/posts/2016-01-01-test.md` makes more sense than `/collections/posts/documents/2016-01-01-test.md`, with the logic being that the posts are the documents, thus it's a duplicative namespace, but it's an easy enough change if you believe otherwise.

mertkahyaoglu:
@benbalter Currently `id` is in `/2016/01/01/test` format. Is there a way for server to change it to `2016-01-01-test.md` ? Or should front-end make the conversion itself?

> I think /collections/posts/2016-01-01-test.md makes more sense

I agree, this one is simple and better.

benbalter:
> Is there a way for server to change it to 2016-01-01-test.md

Yes. Opened https://github.com/jekyll/jekyll-admin/issues/41 to track.

parkr:
> @parkr any clean way to get the pre-rendered HTML with to_liquid? Shouldn't that be the difference between output and content?

[...]

PR #53, opened 2016-07-22
Return raw, unrendered markdown via API

benbalter:
Fixes https://github.com/jekyll/jekyll-admin/issues/35.

My proposed solution here is to add a `raw_content` filed, which will contain the unrendered markdown.

@parkr thoughts?

### DIFF
--- a/lib/jekyll/admin/server/collection.rb
+++ b/lib/jekyll/admin/server/collection.rb
@@ -13,13 +13,13 @@ module Jekyll
 
         get "/:collection_id/documents" do
           ensure_collection
-          json collection.docs.map { |doc| document_without_front_matter_defaults(doc).to_liquid }
+          json collection.docs.map { |doc| document_for_api(doc).to_liquid }
         end
 
         get "/:collection_id/*" do
           ensure_document
           content_type :json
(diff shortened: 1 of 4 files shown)


parkr:
LGTM!

lib/jekyll/admin/server/collection.rb
```diff
@@ -66,20 +66,25 @@ def ensure_document
...
           # If the file has YAML front matter, read it in
           content = File.read(doc.path, Utils.merged_file_read_opts(site, {}))
           if content =~ Jekyll::Document::YAML_FRONT_MATTER_REGEXP
+            content = $POSTMATCH
             data_file = SafeYAML.load(Regexp.last_match(1))
             doc.merge_data!(data_file, :source => "YAML front matter") if data_file
           end
 
+          # Return the unrendered markdown, see https://git.io/vKDTG
+          doc.merge_data!({ "raw_content" => content }, :source => "Jekyll Admin")
```
parkr: We should add this field to `Document` and `Page` in Jekyll proper eventually. So dumb that we don't. `raw_yaml`, too, maybe?

Issue #25, opened 2016-07-05
jekyll/admin or jekyll-admin?


benbalter:
[Technically, technically](http://guides.rubygems.org/name-your-gem/), in Ruby land, `jekyll/admin` would be Jekyll::Admin and `jekyll_admin` would be `JekyllAdmin`.

In Jekyll land, we have a convention of calling everything `jekyll-X` (e.g.., `jekyll-feed`). 

@parkr any strong feelings if this should be `jekyll/admin` or `jekyll-admin`?

parkr:
@benbalter It doesn't make a big difference to me. The only piece is that if you add `jekyll-admin` to `gems` in your `_config.yml`, it will run `require "jekyll-admin"`, so it needs to be require-able from that name. I'd leave it as-is.

nhoizey:
Well, maybe it's time to review/complete this [Jekyll style guide](http://ben.balter.com/jekyll-style-guide/plugins/#naming) and publish it in [the official doc](http://jekyllrb.com/)? ;-)

benbalter:
Ha. Okay. So taking [my own advice](http://ben.balter.com/jekyll-style-guide/plugins/#naming), it should be `Jekyll::JekyllAdmin` so that it can be loaded and distributed as `jekyll-admin` rather than `jekyll/admin`.

nhoizey:
So I wonder, I followed Jekyll Assets example for my [new Jekyll Cloudinary plugin](https://github.com/nhoizey/jekyll-cloudinary), which seems to be a bad idea.

I think such conventions should really be documented and explained.

nhoizey:
@benbalter sorry, but I don't understand why you say this:
> it should be `Jekyll::JekyllAdmin` so that it can be loaded and distributed as `jekyll-admin` rather than `jekyll/admin`

My Cloudinary plugin module is `Jekyll::Cloudinary`, but it is loaded with `gem 'jekyll-cloudinary'` in my `Gemfile`, not `gem 'jekyll/cloudinary'`.

I think I'm missing something here, I'm still new to Ruby and Jekyll plugins…

benbalter:
@nhoizey see http://guides.rubygems.org/name-your-gem/ (which is what I recently learned from).

nhoizey:
Ok, so you recommend not following your advice in http://ben.balter.com/jekyll-style-guide/plugins/#naming ? ;-)

PR #54, opened 2016-07-22
Namespace as JekyllAdmin

benbalter:
This PR renames the project from `Jekyll/Admin` with a namespace of `Jekyll:Admin` to `Jekyll::JekyllAdmin` which is the Jekyll convention, to allow it to be called as `jekyll-admin` like other `jekyll-*` Gems in the user's config. There should be no substantive changes.

/cc @parkr 

@mertkahyaoglu this will move the `public` folder from `/lib/jekyll/admin` to `/lib/jekyll-admin`. If there are any merge conflicts, you can blow away the public folder on this branch and replace it with your own (but Git should be smart enough to resolve the conflict).

Fixes https://github.com/jekyll/jekyll-admin/issues/25.


benbalter:
Took a few tries, but got things merged in. @parkr if I can get a LGTM from you, this should be good to go.

jekyll-admin.gemspec
```diff
@@ -1,11 +1,11 @@
...
 lib = File.expand_path('../lib', __FILE__)
 $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
-require 'jekyll/admin/version'
+require 'jekyll-admin/version'
 
 Gem::Specification.new do |spec|
   spec.name          = "jekyll-admin"
-  spec.version       = Jekyll::Admin::VERSION
+  spec.version       = Jekyll::JekyllAdmin::VERSION
   spec.authors       = ["Ankur Singh"]
```
parkr: This should be updated.

jekyll-admin.gemspec
```diff
@@ -1,11 +1,11 @@
...
   spec.email         = ["ankur13019@iiitd.ac.in"]
```
parkr: Emails too.

lib/jekyll-admin.rb
```diff
@@ -0,0 +1,30 @@
...
+require "jekyll-admin/server/data.rb"
+require "jekyll-admin/server/page.rb"
+require "jekyll-admin/server/static_file.rb"
+
+# Monkey Patche
+require_relative "./jekyll/commands/serve"
+require_relative "./jekyll/convertible_ext"
+
+module Jekyll
+  module JekyllAdmin
```
parkr: If the file is `lib/jekyll-admin.rb`, convention dictates `Jekyll::Admin`. The namespacing here is `lib/jekyll/jekyll_admin.rb`, which seems to get neither of what you're looking for 😉

parkr:
@benbalter Why not just ship as `JekyllAdmin` proper? No need to include anything in the `Jekyll` namespace unless you're monkey-patching our code 😉

benbalter:
@parkr believe I implemented all your feedback.

@mertkahyaoglu how disruptive would it be if we merged this? In practicality, if you had an existing branch, you'd simply need to merge in master and copy whatever's in the `jekyll/admin/public` folder to `jekyll-admin/public` (blowing away whatever's in there). 

If not too disruptive, I'd actually prefer to merge this sooner than later, as it'll likely cause merge conflicts on the Ruby side with just about any subsequent change.

mertkahyaoglu:
@benbalter No problem for me. We can merge this 👍

benbalter:
@mertkahyaoglu let me know if you have any trouble merging master into the open PRs... as I said, it should be safe to delete the contents of the public folder on master and copy in your content to resolve any conflict... it's just a folder rename.

PR #62, opened 2016-07-27
Clean up monkey patching

benbalter:
Since we're just monkey patching Document and Convertible by including a module, we can clean things up a bit, by extending them programmatically, rather than with dedicated, one-line files.

This also moves the `RACK_ENV` default to the JekyllAdmin module, to keep the monkey patch of `commands/serve` as small as possible.


benbalter:
I had been working on this, and @jldec discovered that when including `jekyll-admin` in another site (e.g., non-dev), we got a LoadError as Sinatra thought it was in `:development` mode. 

Pushed f0a1aba which sets the production flag early in the load process, before we load Sinatra.

Issue #65, opened 2016-07-27
Updating document path fails


mertkahyaoglu:
I think this is because server doesn’t expect folder prefix. For example, if the path is `2016-01-02-test2.md`, it creates a file with the requested path but keeps the old one. If it has a folder prefix like this `_posts/2016-01-02-test2.md`, it completely fails. This [line](https://github.com/jekyll/jekyll-admin/blob/master/spec/jekyll-admin/collection_spec.rb#L150) in the test fails to catch the bug. 

/cc @benbalter

PR #66, opened 2016-07-27
Support collection folders being prefixed when renaming documents

benbalter:
This PR fixes https://github.com/jekyll/jekyll-admin/issues/65 and explicitly tests that when a file is renamed, both with path `_posts/2016-01-01-test.md` and path `2016-01-01-test.md`, that (A) the new file is created and (B) the old file is deleted.

To do this, I created a `be_an_existing_file` RSpec matcher, and updated the existing tests to use it in all cases.

Last, in diagnosing this issue, I discovered that both our test fixtures and the API would fail if a folder did not exist. All write operations are now centralized and prefixed by a `mkdir_p` call to ensure the directory exists before we try writing to it (and that it's in the site source).


benbalter:
@tekknolagi Your `static_file_equals` was really elegant, but I opted to refactor it a bit, as part of this PR. When the test failed, since it was just a `.not_to eql(nil)` comparision, it was hard to diagnose what was going on. Hoping to favor readability of tests over conciseness, I moved the assertions to the two individual static file tests, where if they fail, it will now provide a more descriptive message to help diagnose the failed expectation.

mertkahyaoglu:
@benbalter I'm getting `NoMethodError: undefined method delete_file` after 
`curl -X PUT -d '{"meta":{"foo":"baz"},"raw_content": "# test", "path": "page2.md"}' http://localhost:4000/_api/pages/page.md`

benbalter:
@mertkahyaoglu dumb mistake on my part. Fixed via 60184ec.

Issue #82, opened 2016-07-28
Payload for API Requests


benbalter:
Looking at #79, #80, and #81, I think there's some ambiguity in terms of how to describe a document when making requests to and getting responses from the API. At least part of this is my fault, with the late addition of `raw_content` and `front_matter` fields via #61 and #53.

Whatever we do, I think it makes sense for the response and the request to be consistent.

Here's what I'd like to propose as the document (page, post, etc.) payload, and if it sounds good, I'll document it more formally:

* Top level keys are keys with special meaning. This includes:
  * Computed, read-only keys like `url`
  * Computed, read/write keys like `path`
  * Front matter defaults
* The top-level namespace will have `content` and `raw_content` keys with the HTML and markdown respectively
* The top-level namespace will have a `front_matter` key which includes the raw front matter as seen on disk.

A standard page may then look like this:

```json
{  
   "some_front_matter":"default",
   "foo":"bar",
   "content":"<h1 id=\"test-page\">Test Page</h1>\n",
   "dir":"/",
   "name":"page.md",
   "path":"page.md",
   "url":"/page.html",
   "raw_content":"# Test Page\n",
   "front_matter":{  
      "foo":"bar"
   }
}
```

When making a request, clients can set the following top-level fields:

* `raw_content` - the raw, unrendered content to be written to disk (currently `body`)
* `front_matter` - the entire YAML front matter object to be written to disk (currently `meta`)
* `path` - the new file path relative to the site source, if the file is to be renamed

Does that make sense? Anything I'm missing? If so, will change the API to use `raw_content` and `front_matter` rather than `body` and `meta` and will update the docs accordingly.

mertkahyaoglu:
@benbalter It makes sense. With #59, front-end exactly follows this propose. 👍  for `raw_content` and `front_matter`.

PR #59, opened 2016-07-25
Fix updating pages/documents

mertkahyaoglu:
Closes #79 
Closes #80

PR #85, opened 2016-07-28
Standarize on raw_content and front_matter keys

benbalter:
This PR changes the `meta` field to `front_matter` and `body` field to `raw_content` in expected request payloads to match the response payload.

Fixes https://github.com/jekyll/jekyll-admin/issues/82.


benbalter:
@mertkahyaoglu does this match your expectations? Don't want to have you to keep changing the client side code because we keep changing the API... hoping this change can be the last.

mertkahyaoglu:
LGTM. I think we can merge this and finally #59 can be merged afterwards. What do you think?

PR #89, opened 2016-07-29
Fix deleting pages/documents

mertkahyaoglu:
With this PR, front-end can delete pages and documents from listing and editing views properly.


benbalter:
LGTM.

PR #94, opened 2016-07-31
Fixed undefined method error for /_api endpoint and started work on api examples

jakebruemmer:
Started work on improving API documentation with interactive examples at `docs/working_with_the_api.md` and fixed no method error that was being return at the root of the of `/_api` endpoint. `/_api` endpoint now should correctly return a `JSON` response that has information on all of the api endpoints.

## No method error before fix

no_method_error

## JSON response after fix
fixed_no_method_error


benbalter:
@jakebruemmer Thanks for this and sorry for the delay in responding... we weren't quite set up for external contributions when you opened your pull request. I believe the undefined method error was fixed via https://github.com/jekyll/jekyll-admin/commit/9ce25a79d097511e3ffb92c58d59f716bfa5dc85.

I think there's still value in the documentation though. What do you think about making the examples cURL requests so that they are language agnostic?

jakebruemmer:
@benbalter I will look into doing that if that's still something that you think would be valuable. I've been caught up at work recently and haven't had much time. Do you know if there's anyone else that's working/has been working on this in the mean time?

PR #101, opened 2016-08-03
Fix searching documents/pages

### LINKED ISSUE #100: Search component disappears after a search with no results.

mertkahyaoglu:
Fixes #100 

Additionally, I made the search mechanism more generic. Now every reducer implements their own filter selector rather using the one in utility reducer. In this way, they can choose the field to be filtered by. This will allow us to search static and data files by filename.


benbalter:
>  I made the search mechanism more generic. Now every reducer implements their own filter selector rather using the one in utility reducer. In this way, they can choose the field to be filtered by. This will allow us to search static and data files by filename.

Nice. LGTM.

PR #102, opened 2016-08-03
Sort documents by date

### LINKED ISSUE #95: Show post date in post index list
It'd be nice to show the date of posts for the post list (and to sort reverse chronologically by date, which we can do on the API level, if you want).

Other collections could theoretically have dates, but there's no programatic way to tell, so I think it's fine just to do this for posts for now.

mertkahyaoglu:
Closes #95


benbalter:
LGTM. Is that going to be ascending or descending? (I think it should be latest first)

mertkahyaoglu:
Yes, it is latest first.

mertkahyaoglu:
This works for only post documents because other collections' documents have the same date which is the start date of the server. I think we can fix that later on.

benbalter:
> This works for only post documents because other collections' documents have the same date which is the start date of the server. I think we can fix that later on.

👍

PR #103, opened 2016-08-04
Front-end: Data files

### LINKED ISSUE #73: Data file support
- [x] Implement backend API
- [ ] Ability to edit `_data` files

mertkahyaoglu:
- [x] List
- [x] Search
- [x] Delete
- [x] Create
- [x] Edit
- [x] Tests

Closes #73


mertkahyaoglu:
@benbalter May I borrow your 👀  ? 😄 

The only missing part is the form validation. Right now, leaving filename and content empty is not being checked. I passed that for this PR because I'm going create another one that makes validation system more general. Currently, [it](https://github.com/jekyll/jekyll-admin/blob/master/lib/jekyll-admin/public/src/utils/helpers.js#L168) is not applicable to any other content type and looks bad 👎

DirtyF:
@mertkahyaoglu great job, it's doing the job. I could edit, create and delete a data file.

A small feedback regardintg the UI:  

* breadcrumbs are not present when editing a data file like for other documents, it would be more consistent to add it there too.
* when a search doesn't return any result, the message is "You don't have any data file", I would rather suggest something like "No data file named SEARCH TERM , would you like to create it ?"

mertkahyaoglu:
@DirtyF Thanks for reviewing this ❤️ 

> breadcrumbs are not present

Yep, I should better add breadcrumbs.

> No data file named SEARCH TERM , would you like to create it ?

I'm not sure about this because `SEARCH_TERM` don't have to be a filename. For example, users might want to search all of the posts that their filename contains `bird` or all of the posts with the prefix of `new`. So creating them makes no sense but yeah, we could improve the message like this, `You don't have any data file with SEARCH_TERM`

One thing that's in my mind not for searching but listing, when there are no posts/pages available, we could show a button to create the first one.

benbalter:
> I passed that for this PR because I'm going create another one that makes validation system more general.

👍 Agree 1. get it working, 2. add polish, especially for an initial release. Worst case scenario, the API will return an error, which hopefully the front end could at least guard against and say e.g., "something went wrong, please try again"

benbalter:
Nice work! LGTM.

I'll open a separate issue as a follow up (feel free to merge this as is, if you're happy with it), but two edge cases:

1. Data files can have a `YAML` extension
2. Data files can be JSON (front end should likely behave the same, as YAML'd easier to edit, but API can [save back to JSON based on the extension](https://github.com/jekyll/jekyll-admin/issues/106))

mertkahyaoglu:
@benbalter Should I wait for #118 to merge this or another PR for that?

benbalter:
LGTM.

@mertkahyaoglu I'd say lets merge this, so you can continue to make progress, and we can change the API/data model in a subsequent PR.

Issue #104, opened 2016-08-05
Server returning `null` after creating a data file with an extension


mertkahyaoglu:

benbalter:
@mertkahyaoglu sorry about that. Do you have an example cURL that can recreate this? We have a test in place that ensures the response, but digging in now.

mertkahyaoglu:
Here it is `curl -X PUT -d '{"test":"test"}' http://localhost:4000/_api/data/test10.yml`

mertkahyaoglu:
@benbalter Ah, got it now, it should not have a file extension 😬

benbalter:
Got it. Yep, the issue is the `.yml` extension, which means the server isn't following the spec. I'll get a fix up momentarily, as you should be passing the extension in the event it's a JSON file, which is also supported.

mertkahyaoglu:
Yeah, but it creates a data file like test10.yml.yml

benbalter:
https://github.com/jekyll/jekyll-admin/pull/107 Should allow the server to accept data files with and without an extension.

mertkahyaoglu:
Nice 👍

Issue #106, opened 2016-08-05
Ability to create/update JSON data files


benbalter:
e.g., `_data/foo.json` versus `_data/foo.yml`. Based on the extension, we'll want to call `to_json`, rather than `to_yaml`.

benbalter:
To note, blocked by https://github.com/jekyll/jekyll-admin/pull/107.

PR #107, opened 2016-08-05
Data files should default to `.yml` extension

benbalter:
This pull request makes it so requests to `/data/*` in the API default to the `.yml` extension, when none is given, but also supports accepting an explicit extension.

For now, you can `GET` and `DELETE` files with a `.json` extension. I've opened https://github.com/jekyll/jekyll-admin/issues/106 to track writing JSON data files, but that's not critical for an initial release / GSoC in my mind.

Fixes https://github.com/jekyll/jekyll-admin/issues/104


benbalter:
👍 'd via https://github.com/jekyll/jekyll-admin/issues/106.

PR #109, opened 2016-08-05
Refactor data file handling

benbalter:
A better fix for #104, and building on #107 this fixes an issue where `File.extname` would return `.yml` instead of `yml` resulting in `_data/file..yml`. In fixes this, I broke out `data_file_name` to represent the data file's key, and `data_file_extension` to represent it's file type.


mertkahyaoglu:
LGTM 😬

PR #115, opened 2016-08-08
Add a disclaimer for the current installation process

### LINKED ISSUE #113: Could not find a valid gem
I get this error while installing jekyll-admin using `gem`:
```shell
ERROR:  Could not find a valid gem 'jekyll-admin' (>= 0) in any repository
```
I already add it to the `Gemfile`, and the `_config.yml`. Is there something that I'm doing wrong?

DirtyF:
Write explicitly in the README that installation does not involve a gem yet. (to delete after the first gem release)

Temporary fix #113

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,8 @@ A Jekyll plugin that provides users with a traditional CMS-style graphical inter
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


benbalter:
Nice clarification. Thanks @DirtyF!

Issue #110, opened 2016-08-06
API Datafile PUT response


mertkahyaoglu:
Right now, after a PUT request `curl -X PUT -d '{"test":"test"}'`, server responds with;

```
{"test":"test"}
```

Is it possible to include the filename here ? Like this;

```
{
  "file_name": {
     "test":"test"
  }
}
```

So users are redirected to the created datafile page looking to the response filename.

benbalter:
Data files aren't currently objects in Jekyll (see https://github.com/jekyll/jekyll/issues/5207), but we can implement that behavior ourselves.

benbalter:
@mertkahyaoglu Can you explain a bit more what you're trying to solve for? If you're making a `PUT` request to `/data/filename.yml`, a get request to that same endpoint would return the file (or is it that you want the object to know it's own filename so it can be passed around more easily?)

mertkahyaoglu:
> you want the object to know it's own filename so it can be passed around more easily

That was my aim. I got a messy solution but it's OK for now.

benbalter:
Slightly larger technical lift because it's not a native object in Jekyll land, but have a solution in the works.

PR #118, opened 2016-08-08
Make data files their own object

benbalter:
Fixes https://github.com/jekyll/jekyll-admin/issues/110.

This PR introduces a new object called a DataFile, which is used to generate hashes with metadata for data files. Here's an examples payload:

```json
{
  "path": "_data/data_file.yml",
  "relative_path": "_data/data_file.yml",
  "slug": "data_file",
  "ext": ".yml",
  "title": "Data File",
  "raw_content": "foo: bar\n",
  "content": {
    "foo": "bar"
  }
}
```

I tried to follow the Document payload structure as closely as possible, with `raw_content` representing the content on disk and `content` representing the parsed / final content.

@mertkahyaoglu does this meet your needs?

If so, I believe I just need to update the *request* payload, but this should be good to go.


mertkahyaoglu:
This is awesome 🎉

benbalter:
@mertkahyaoglu updated this PR (and the docs) to reflect the updated API spec. You can now pass:

* `content` - The JSON encoded YAML object to write to disk as the file's content
* `raw_content` - The raw string representing the file's content to write to disk

Both will write to disk as YAML.

After this PR merges, the extension will be optional, defaulting to `.yml`.

Since this is a breaking change, feel free to merge this PR at your convenience (or let me know if you need other changes).

LGTM.

benbalter:
LGTM.

PR #120, opened 2016-08-10
Update static file API to expect base64 encoded files

benbalter:
This design pattern is based off [the GitHub Content API](https://developer.github.com/v3/repos/contents/#create-a-file) and expects that raw files will be submitted as base64 encoded strings, and returned as the same to get around binary and encoding issues.

Request payloads may now submit `raw_content` or `encoded_content` fields, both of which will be written to disk (with `encoded_content` being base64 decoded first). You can always use `encoded_content`, but could use `raw_content` for content which you were confident did not have any potential encoding issues that might break JSON.

The file is always returned with the `encoded_content` field, and will not have `content`, `raw_content` or `front_matter`.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -51,7 +51,7 @@ Data files and the config file are direct JSON representations of the underlying
 
 #### Static files
 
-Static files are just the raw content and may be binary.
+Static files are non-Jekyll files and may be binary or text.
 
 ### Collections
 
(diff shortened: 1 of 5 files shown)


benbalter:
@mertkahyaoglu if this spec looks good to you, please feel free to cherry-pick or merge as is convenient for your work.

LGTM.

mertkahyaoglu:
Let's merge this.

PR #122, opened 2016-08-11
Front-end: Static files

### LINKED ISSUE #72: Static file support
- [x] Implement backend API
- [ ] Ability to edit list and upload static files

mertkahyaoglu:
- [x] List
- [x] Search
- [x] Upload
- [x] Delete
- [x] Tests

Closes #72


mertkahyaoglu:
@benbalter I think this is ready to go. Would love to know your thoughts 👀

benbalter:
@mertkahyaoglu this is awesome and LGTM. Nice work!

A few minor suggestions to polish things up a bit:

1. It'd be cool if clicking the file would let you view the file (e.g., I want to see the image). Opening the file in browser, in a new tab should be sufficient in my mind. The API should give you the path.

2. It's strange that the "upload files" link isn't a button like the other content types (pages, posts, etc.) are. I thin the icon and text are fine, but I'd suggest keeping it as a button for visual consistency... it helps it stand out and took me an extra second to find because I was expecting a big orange button. 

3. After I upload a file you should either (A) sort the list, or (B) re-call the index so that documents are sorted.

4. I was able to upload the same document twice, which isn't terrible, but it appeared twice on the document list until I reloaded or moved to a different page and back. Again, I'd either re-call the index (which probably makes more sense) after an upload, or ensure uniqueness of paths (for when a file is overwritten).

5. If a file is being overwritten, what do you think about a warning, similar to when a file is deleted?

6. I'd add some help text to hint to users that the page is a drop zone (nicely done!)

7. Super minor, but do we really need to versions of the Jekyll logo in the fixture site? Feel free to delete one or the other and update the path in tests accordingly.

mertkahyaoglu:
@benbalter  I implemented your suggestions. Could you check this out?

The only problem is that we still cannot view the uploaded files in development mode because of the mismatching port numbers. It should be working in production.

benbalter:
LGTM. Nice work @mertkahyaoglu!

PR #124, opened 2016-08-14
fix link to development docs

DirtyF:
Don't know if you prefer to have a valid link on Github or locally on Jekyll ?


benbalter:
@DirtyF if you'd like, you can link to http://jekyll.github.io/jekyll-admin/development/, which will work in both?

benbalter:
LGTM. Thanks, as always, @DirtyF!

PR #125, opened 2016-08-15
Follow API changes for Data Files

### LINKED ISSUE #114: Support for data files with a YAML extension
Same behavior as `.yml`, but should be able to create/delete/edit `.yaml` files as well.

I believe the API should treat them both the same.

### LINKED ISSUE #116: Support for JSON-based data files
I suspect we'd want the editor to continue to be YAML, since it's easier for humans to work with, but on save, the API would save the file as JSON, rather than YAML. This issue is to track the front end implementation. See https://github.com/jekyll/jekyll-admin/issues/106 for the backend side of things,.

mertkahyaoglu:
Closes #116 and closes #114


benbalter:
Nice work. LGTM.

PR #129, opened 2016-08-17
Do not add empty metadata fields

### LINKED ISSUE #81: Don't save empty metadata fields
Steps to reproduce:

1. Navigate to http://localhost:4000/admin/pages/page.md
2. Edit a metadata field
3. Hit save

Expected:

No new fields added

Actual:

"New field 1", "New field 2", "New field 3", etc. added to the front matter, even if they have no content.

mertkahyaoglu:
Fixes #81

### DIFF
--- a/lib/jekyll-admin/public/src/actions/collections.js
+++ b/lib/jekyll-admin/public/src/actions/collections.js
@@ -92,7 +92,11 @@ export function putDocument(id, collection) {
     dispatch({type: ActionTypes.CLEAR_ERRORS});
     const doc = JSON.stringify({
       path,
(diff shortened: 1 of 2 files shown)


benbalter:
This LGTM. 

My only question, is would there ever be a time when you wanted a metadata field that was an empty string? I can't think of one, but asking outloud in case I'm missing one.

mertkahyaoglu:
I can't think of one neither. Let see if the community wants that.

BTW, what's wrong with the jekyllbot ? 😁

benbalter:
LGTM!

PR #130, opened 2016-08-17
Move frontend to root folder

mertkahyaoglu:


mertkahyaoglu:
@benbalter Do we need `script/return-to-pwd` anymore?

benbalter:
> Do we need script/return-to-pwd anymore?

The build script still needs to `cd` into `src`, right?

I'm not tied to it... my motivation was that running `script/server`, `script/build`, `script/cibuild`, etc. shouldn't leave you in a different folder than when you started... if that's not the case any more, and it's not necessary, by all means, 💀 it.

mertkahyaoglu:
> The build script still needs to cd into src, right?

Ops, forgot that.

mertkahyaoglu:
> The build script still needs to cd into src, right?

No, it does not. `package.json` in the root now.

mertkahyaoglu:
@benbalter is it cache that fails travis?

mertkahyaoglu:
@benbalter Anything else you want me to add?

benbalter:
I suspect we want to `.gitignore` the compiled/built site? I'll follow up with another PR that builds the site prior to releasing, otherwise,

LGTM.

PR #131, opened 2016-08-17
Add script/release

benbalter:
This adds a quick `script/release` script, the idea being, that if we get in the habit of using this, and it automatically builds the site, we won't try to release the gem without the latest compiled site.

This also adds the built site to `gitignore` to keep diffs cleaner (and not having to ask contributors to constantly run `script/build` prior to contributing).


mattr-:
LGTM

benbalter:
Thanks for the :eyes: @mattr-!

PR #135, opened 2016-08-18
Slightly better messages when search fail

DirtyF:
When search do not return anything, it feels odd to read "You don't have any pages." 
Tried to make it more accurate with this proposition.

### DIFF
--- a/src/containers/views/DataFiles.js
+++ b/src/containers/views/DataFiles.js
@@ -95,7 +95,7 @@ export class DataFiles extends Component {
           files.length > 0 && this.renderTable()
         }
         {
-          !files.length && <h1>{`No matching data files found.`}</h1>
+          !files.length && <h1>{`No data files found.`}</h1>
         }
       </div>
     );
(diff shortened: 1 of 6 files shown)


benbalter:
LGTM. Thanks @DirtyF! @mertkahyaoglu what do you think?

mertkahyaoglu:
Let's say, there are not any pages on the site. We are going to get `No matching pages found` at the beginning, even though we are not searching anything. The message should be specific to only searching. We need to find a way to separate them.

Crunch09:
Maybe `No XYZ found` instead of `No matching XYZ found`? I think this would work well in both cases.

mertkahyaoglu:
Adding this might work;
``` javascript
{
    !filterByFilename(files, utils.input).length && <h1>{`No matching data files found.`}</h1>
}
```

mertkahyaoglu:
@Crunch09 That also works.

DirtyF:
OK, thanks for the explanation @mertkahyaoglu, I'll update with @Crunch09 proposition for now.

mertkahyaoglu:
Thanks @DirtyF 👍

Crunch09:
@DirtyF  Thanks for the mention 😆

benbalter:
@mertkahyaoglu if this looks good to you, feel free to merge. :smile:

mertkahyaoglu:
LGTM

Issue #63, opened 2016-07-27
Document indexes shouldn't include document content


benbalter:
Otherwise, on a large site, with a lot of posts, the `/posts/` endpoint would be extremely large.

benbalter:
@mertkahyaoglu would this cause a problem if the document index didn't return the `content` or `raw_content` fields? I assume an individual page view calls the individual page's endpoint?

mertkahyaoglu:
> I assume an individual page view calls the individual page's endpoint

Yes, that's correct. We don't need them other than individuals' endpoints.

PR #138, opened 2016-08-18
Remove file contents from all indexes

benbalter:
This PR removes the `content`, `raw_content, `encoded_content`, and `docs` keys from all index API responses.

To do this, `JekyllAdmin::DataFile` and `Jekyll::Collection` now include `Apiable`, such that all API responses now go through `Apiable#to_api` (which itself is based on the class's native `to_liquid` method).

`to_api` now also accepts an `include_content` argument, which, you guessed it, includes the content in the response.

Fixes https://github.com/jekyll/jekyll-admin/issues/63


benbalter:
@mertkahyaoglu it looks like the static files index uses the encoded content field to display image previews. Since this PR removes the encoded content field from the index response, would you mind please:

1. Update the front end to load the content from the Jekyll server directly
2. Update the front end to load the content from the individual file's endpoint?

screen shot 2016-08-19 at 10 53 57 am

If you can add that fix to this PR, we can merge the two together. Hope it's a simple fix.

mertkahyaoglu:
> Update the front end to load the content from the Jekyll server directly

I didn't understand what you mean loading from the Jekyll server directly. It already does that.

> Update the front end to load the content from the individual file's endpoint?

So for the current view which lists all static files with their content, front end first calls the index to retrieve the list of static files and then calls individual endpoints for each one, right? Or are you suggesting it should list them with only filenames and to be able to see a file content, users should visit the individual view?

benbalter:
@mertkahyaoglu based on that error message in console (base64) I thought it was trying to pull the content from the index as a base64 encoded data string. If that's the case, we need to either:

1. Have the front end call the individual endpoint for images to get the content
2. Have the browser call the actual file directly (e.g., ``

mertkahyaoglu:
Hmm, I get it now.

> Have the browser call the actual file directly (e.g., ``

This makes sense. Why didn't I think of that ? 😬  It's an easy fix. The only problem we won't be able to see it for development cause of the port number. (It will try `https://localhost:3000/image.pmg`)

mertkahyaoglu:
I'm going to create a PR about the port number issue. Simply will look at the `process.env.NODE_ENV` variable to see if `development` or `production`. If `development`, it will add prefix `http:localhost:3000`, otherwise the image source will be just `/image.png`.

This is also needed for previewing files in development mode.

benbalter:
@mertkahyaoglu so as far as the front end goes, this PR should be good to merge?

mertkahyaoglu:
I'm now creating a PR about notification system (so much changes). Can we merge this after that? Then I can fix the frontend in another PR.

mertkahyaoglu:
LGTM :shipit:

PR #147, opened 2016-08-19
Load static files directly from Jekyll server

mertkahyaoglu:
Front-end was using `encoded_content` for previewing static files. Since it only lists all of the static files and index API responses include file contents, there was no need to call individual endpoints. With #138 , file contents will be removed from all indexes so `encoded_content` will not be available for index API. With this PR, front-end loads static files directly from Jekyll server using `path` field without needing the actual contents.

### DIFF
--- a/src/components/FilePreview.js
+++ b/src/components/FilePreview.js
@@ -15,11 +15,15 @@ export default class FilePreview extends Component {
     const { file } = this.props;
     const extension = file.extname.substring(1);
     const image = /png|jpg|gif|jpeg/.test(extension);
+    let prefix = '';
+    if (process.env.NODE_ENV == 'development') {
+      prefix = 'http://localhost:4000';
+    }
     return (
       <div className="file-preview">
-        <a href={file.path} target="_blank">
+        <a href={prefix+file.path} target="_blank">
         {
-          image && <img src={`data:image/${extension};base64,${file.encoded_content}`} />
+          image && <img src={prefix+file.path} />
         }
         {
           !image &&


src/components/FilePreview.js
```diff
@@ -15,11 +15,15 @@ export default class FilePreview extends Component {
     const { file } = this.props;
     const extension = file.extname.substring(1);
     const image = /png|jpg|gif|jpeg/.test(extension);
+    let prefix = '';
+    if (process.env.NODE_ENV == 'development') {
+      prefix = 'http://localhost:4000';
```
benbalter: Shouldn't this be `3000`?

benbalter:
One comment, otherwise, LGTM.

mertkahyaoglu:
@benbalter I just noticed the problem and was commenting it. It should be the port that the real server is running on so it can retrieve the files from Jekyll server properly (There is no `http://localhost:3000/image.png`). For development mode, front-end does not really need to know which port the real server is running in order to call the API because it just uses `/_api`. But when it comes to previewing it must know the full path. 

Above I just assumed that the server was running on `4000` which is the default but it can change. What can we do about this? Can we add the running port to `_api` endpoint.

mertkahyaoglu:
I'm thinking two solution for this;

1. server keeps the port number as an environment variable and front-end reads that variable. 
2. front-end reads the port number from the configuration if it is set, otherwise uses `4000`.

benbalter:
@mertkahyaoglu misread the code. You are correct (was thinking the front-end port, not the backend port). I think for an initial release, it's fine to assume the server is always running on `4000`, as I suspect that's a very strong default. Since it's development, it's also fine to say it needs to be running on `4000` for development, and any more short work in production/real world use.

PR #148, opened 2016-08-20
Added default date value to post titles

codegaze:
I believe most of the times someone creates a new post needs the current date. 
Maybe the post title should have a default value of "YYYY-MM-DD-your-title.md" instead of a plain placeholder.

This just  adds the current date for posts only if there is no 'path' value with the use of moment.js.

Just a thought :)


benbalter:
@mertkahyaoglu what do you think?

mertkahyaoglu:
❤️  that, LGTM

mertkahyaoglu:
@codegaze If you could add a test for this, it would be awesome and we can merge afterwards. Glad to help with the test If you need any 👍

codegaze:
Thank you for your patience @mertkahyaoglu  :smiley: I 'll get on it today.

If I understand correctly I need to create a new const 'post_content' in fixtures with no 'path' and test the 'defaultValue' to the current date?

mertkahyaoglu:
@codegaze That's an option. You can also directly pass an object to `setup` function like this;

``` javascript
setup({ link:'test', type:'posts', path:'' })
```

And it would be better to test `placeholder` as well.

codegaze:
@mertkahyaoglu I added a test, please check if everything is ok. Hopefully I didn't mess anything with my git experiments. :flushed: 
I can help with the front end if you need more people but you have to endure my git issues :sweat_smile:

mertkahyaoglu:
Wooah 😁  You always need to check that your branch is up to date with master before committing.
You can use the commands below to achieve that;

1. `git fetch origin`
2. `git merge origin/master`

or you can use the `Update branch` button above.

You can close this and create a fresh PR if you'd like.

PR #149, opened 2016-08-20
Windows mkdir fix

codegaze:
`mkdir ./lib/jekyll-admin/public` will fail in a windows machine, at least for mine.
If someone else is having the same issue and it's not my configuration please add this.

Added [mkdirp ](https://github.com/substack/node-mkdirp) devDependency and a new script operation `create-dist` to create the 'public' directory without problems.


package.json
```diff
@@ -10,10 +10,11 @@
     "open:src": "babel-node tools/srcServer.js",
     "lint": "esw webpack.config.* src tools",
     "lint:watch": "npm run lint -- --watch",
-    "clean-dist": "npm run remove-dist && mkdir ./lib/jekyll-admin/public",
+    "clean-dist": "npm run remove-dist  && npm run add-dist",
+    "add-dist": "mkdirp ./lib/jekyll-admin/public",
     "remove-dist": "rimraf ./lib/jekyll-admin/public",
     "build:html": "babel-node tools/buildHtml.js",
-    "prebuild": "npm run clean-dist && npm run build:html && npm run lint && npm test",
```
mertkahyaoglu: I think we do need to lint and test before building just to double check everything as expected.

mertkahyaoglu:
@codegaze Thanks for the PR. 

I haven't tested Jekyll Admin in Windows yet. Let's see others are having the same issue as well. Tested this PR locally and no break for mac at least 😬 

Little tip: you might want to do `git branch -f new-branch master` if you have another branch which is not merged yet so commits from that branch will not show up here.

package.json
codegaze: You are right. I really don't understand why my version is different. :|

codegaze:
Thanks for the tip @mertkahyaoglu :smiley:

benbalter:
This looks like this only affects local windows development? It looks like there's a stray commit in there, otherwise I think this can be merged?

mertkahyaoglu:
@benbalter Agree. Let's merge this once you clear that commit @codegaze  👍

codegaze:
@mertkahyaoglu @benbalter No prob, but I need some help to clear the commit, if you have the time to do it of course. :flushed:.

benbalter:
Thanks @codegaze! Cherry picked the two relevant commits via https://github.com/jekyll/jekyll-admin/pull/170 and confirmed things build on windows (and test that they continue to do so going forward).

PR #153, opened 2016-08-20
WIP: Windows tests

benbalter:
This pull requests adds Appveyor Windows tests, and fixes a bug whereby collections, including posts, would not work on Windows due to improper path sanitization.


mertkahyaoglu:
Awesome 👍🏻

benbalter:
Got tests passing. For some reason Appveyor doesn't like rubocop, which I disabled, since it'll run on Travis. Going to merge, since this fixes Windows support, but not going to make this a required test, as from my experience, Appveyor can sometimes be flaky.

To note, the setup is similar to travis, in that it uses `script/bootstrap` and `script/server` and runs the two test suites (Node, Ruby) in parallel.

PR #154, opened 2016-08-20
Exclude next/previous contents from indexes

benbalter:
A follow up to https://github.com/jekyll/jekyll-admin/pull/138, #138 removed the `content` fields from documents when retrieving an index... but we still included the content of the next/previous document, and did so recursively. 

So while the document itself had no content, for each document we loaded, we loaded up to four document's contents in the next/previous fields.

This PR updates the `to_api` logic to not include the content of nor the next/previous document for other documents when generating a given document's next/previous fields.

### Before

```json
{
  "path": "whatever",
  "next" {
    "path": "something",
    "content": "foo",
    "next": { 
     ...
    } 
    ...
  }
}
```

### After

```json
{
  "path": "whatever",
  "next" {
    "path": "something"
    ...
  }
  ...
}
```

I don't believe next/previous was used by the front end, and thus no front end change should be necessary. This does however, make document lists load significantly faster on large sites.

### DIFF
--- a/lib/jekyll-admin/apiable.rb
+++ b/lib/jekyll-admin/apiable.rb
@@ -62,7 +62,7 @@ module JekyllAdmin
     end
 
     def file_contents
-      @file_contents ||= File.read(file_path, file_read_options)
+      @file_contents ||= File.read(file_path, file_read_options) if file_exists?
     end
 
     def file_read_options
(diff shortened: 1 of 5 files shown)


mertkahyaoglu:
👍🏻

PR #155, opened 2016-08-20
Fix datetime picker

### LINKED ISSUE #142: Datetime picker appears hardcoded to the 8th minute
## Reproduction steps:

1. add a new `date` field
2. click on the date picker
3. pick any date

## Expected result
The datepicker should set the time to the current time 

## Actual result
The datepicker appears to use the current 30-minute window but appears to hardcode the last digit of minutes to the value `8`. Even when you manually edit the time in the field itself, it changes back to `8` when using the picker to change the 30-minute increment.

Here's a short video demo: http://link.olivierlacan.com/0p3B0S0N2h2Z

mertkahyaoglu:
Fixes #142


benbalter:
Didn't test this locally, but if you did, LGTM.

PR #159, opened 2016-08-21
Fix Safari datetime parsing

mertkahyaoglu:
Safari somehow does not support `YYYY-MM-DD hh:mm:ss` date format and gives a parse error when you try to visit the document listing page or any other page which a date parsing occurs.

This PR deletes the `dateformat` package and uses already installed `moment` package instead.
Now it parses dates properly in Safari.


benbalter:
LGTM, if you want to get this in before the initial release.

mertkahyaoglu:
There is still some issue with showing date in metadata field on Safari. It may not be ready for the initial release.

benbalter:
Tested locally, and this fixes the issue with Posts not showing in Safari.

Issue #164, opened 2016-08-22
`baseurl` not applied to preview links in Jekyll Admin


mmistakes:
When using `baseurl` in `_config.yml` the preview links in Jekyll Admin don't seem to apply it to the path causing 404's.

### Steps to reproduce:

1. `jekyll new admin-test`
2. Add `gem 'jekyll-admin', group: :jekyll_plugins` to Gemfile
3. `bundle install`
4. Add `baseurl: "/blog"` to _config.yml
5. `bundle exec jekyll serve`
6. Open `http://localhost:4000/admin`, click Posts, and then preview a post (clicking eye icon) or clicking the View button.

### Expected:

`http://localhost:4000/blog/<post-url>/` instead of `http://localhost:4000/<post-url>/`. This applies to pages as well.

mertkahyaoglu:
@benbalter Should we add the `baseurl` prefix to the `url` in the front end if it is present in the `_.config.yml` or should the server handle that and convert all of the `url`s accordingly?

benbalter:
Ohhh Jekyll. While you an I would call `url` something that starts with `http`, the problem is that `url` in Jekyll parlance is actually the HTML path to the site's base url (with `path` being the path on disk), both coming from `to_liquid`, and matching Jekyll's internal data structure prior to that.

I'm going to suggest adding an `html_url` (or `http_url`) field (or something more creatively named), which would be the full URL, with port, scheme, host, path, etc. (which could also parallel the `api_url` via https://github.com/jekyll/jekyll-admin/issues/37).

mertkahyaoglu:
Thanks @mmistakes This is closed via #166, #167

PR #166, opened 2016-08-22
Create new URLable module to include API URL and HTTP URLs in each response

benbalter:
This creates a new URLable module, that gets included in all response classes to provide `http_url` and `api_url` fields (and tests to ensure they're valid for all associated resources).

* `http_url` - Absolute URL to the HTTP(S) rendered/served representation of this resource (e.g., what you're put in your browser to see the thing.

* `api_url` - Absolute URL to the API representation of this resource (e.g., the API endpoint for that thing).

This is a partial fix for https://github.com/jekyll/jekyll-admin/issues/164, with the idea being that the front end would use `http_url` (the absolute URL) rather than `url` (Jekyll's internal representation of the path to the object relative to the base url) when linking to resources. 

This PR should also resolve port issues from the test server, as the URL contains the local serve's port, as retrieved from the site config.


mertkahyaoglu:
Tested locally and it works great. What about excluding `url` ? I guess we don't need it anymore.

benbalter:
> What about excluding url ? I guess we don't need it anymore.

:metal: great idea. I would suggest doing that in a subsequent PR, as that should be a breaking change in terms of the front end (which I believe relies on the URL field today.

If I merge this now, you can update the front end at your convenience, and then we can remove the URL field (or you can do so in your PR by following the examples in APIable), and `master` should continue to be stable.

mertkahyaoglu:
👍

mertkahyaoglu:
I think we cannot get rid of that `url`. What will we use for documents in an unwritten collection like `puppies` ?

benbalter:
> What will we use for documents in an unwritten collection like puppies ?

What do you mean? For the path?

mertkahyaoglu:
I mean if you check `http://localhost:4000/_api/collections/puppies/rover.md`, `http_url` is null for  that document. I didn't get why is that for documents which their collection is not writable, Implemented [here](https://github.com/jekyll/jekyll-admin/blob/master/lib/jekyll-admin/urlable.rb#L9).

mertkahyaoglu:
@benbalter Ah, sorry for the silly question. What was I thinking? Yeah, we shouldn't have `http_url` for the documents that don't have html in _site folder. I didn't notice that for `puppies` that till now 😬 .

The problem is how does the front-end now that documents in a collection is not writable. Because all documents use the same component which relies on the `http_url` field for previewing. If some documents do not have it, front-end crashes.

mertkahyaoglu:
And I think I found a solution for that. Now creating a PR.

benbalter:
> If some documents do not have it, front-end crashes.

Hmm. Gotcha. Is it possible to hide the preview button if `http_url` is null? I don't know what URL we'd show for data files or collections that don't render otherwise.

mertkahyaoglu:
Yep. Just implemented that with #167

PR #167, opened 2016-08-22
Use http_url for preview links

mertkahyaoglu:


src/actions/tests/fixtures/index.js
```diff
@@ -90,7 +90,7 @@ export const page = {
   name: "contact.md",
   raw_content: "# This is the base Jekyll theme.",
   dir: "/",
-  url:"/page.html",
+  http_url:"/page.html",
```
benbalter: In the real API, this would be `http://localhost:4000/page.html`.

src/containers/views/tests/fixtures/index.js
```diff
@@ -13,7 +13,7 @@ export const doc = {
   title: "The Revenant",
   path: "the-revenant.md",
   draft: false,
-  url: '/the-revenant.html',
+  http_url: '/the-revenant.html',
```
benbalter: Same here and line 27 with the scheme + host + port, if it matters.

mertkahyaoglu: Doesn't matter for now, but better fix them for the future 👍

PR #170, opened 2016-08-23
Build the front end in CI

benbalter:
To detect things like https://github.com/jekyll/jekyll-admin/pull/149, this instructs CI to build the front end prior to running tests, only in CI.

Both Appveyor and Travis set the CI environmental variable (which won't be set locally).


script/cibuild-node
```diff
@@ -3,4 +3,11 @@
...
 
 echo "Running Node tests..."
+
+# We want to run script/build on CI to ensure things actually build, but
+# we don't want to have to run script/build locally every time we run tests
+if [ -n "$CI" ]; then
+  script/build
+fi
+
 npm test
```
mertkahyaoglu: We shouldn't need `npm test` here. I believe `npm run build` runs test before building.

benbalter:
Cherry picked @codegaze's two commits from https://github.com/jekyll/jekyll-admin/pull/149 in hopes that they fix CI on Windows.

PR #177, opened 2016-08-24
Add issue template

benbalter:
After a few issues required a bit of back and forth to clarify the reporter's intentions, I thought it'd be best to add a lightweight issue template to help encourage higher quality issue reporting going forward.

Obviously reporters are free to ignore the template, and we want to keep it light because we want people to report issues and submit feature requests, but I'm hoping this can provide a bit more clarify around what's being asked/reported.


mertkahyaoglu:
👍

PR #182, opened 2016-08-24
Added inactive state class to save button

codegaze:
This adds a 'btn-inactive' class to all save buttons on inactive state. I believe the inactive state isn't really straightforward to the user right now, and having an inactive class that can be used anytime is a plus imo :)

Changed the mixin to accept more parameters with default values like  $cursor: pointer, $opacity: 1, $hover: true in order to avoid losing consistency. 

Hope you like it :smiley_cat:

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -77,6 +77,10 @@ i {
   @include btn($btn-active, $btn-active-border);
 }
 
+.btn-inactive {
+  @include btn($btn-inactive, $btn-inactive-border, default, .4, false); 
+}
+
 .btn-success {
   @include btn($btn-success, $btn-success-border);
 }
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -78,7 +78,7 @@ i {
 }
 
 .btn-inactive {
-  @include btn($btn-inactive, $btn-inactive-border, default, .4, false); 
+  @include btn($btn-inactive, $btn-inactive-border, default, .5, false); 
 }
 
 .btn-success {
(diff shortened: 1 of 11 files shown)


src/styles/main.scss
```diff
@@ -77,6 +77,10 @@ i {
   @include btn($btn-active, $btn-active-border);
 }
 
+.btn-inactive {
+  @include btn($btn-inactive, $btn-inactive-border, default, .4, false);
```
mertkahyaoglu: What about `.5` for opacity? It's hard to see the text on the button. At least for my computer 😬

mertkahyaoglu:
Thanks @codegaze I liked the idea of separating inactive state 👍

mertkahyaoglu:
Can we apply the same style for the button in Configuration view?

codegaze:
Missed that!
Morning + :coffee: + The Class and we are ready :wink:

codegaze:
Now we are ready!
@mertkahyaoglu Do you think the save button for save/edit could be abstracted to its own component with safety?

mertkahyaoglu:
Just one comment and ready to merge!

> Do you think the save button for save/edit could be abstracted to its own component?

Yeah, this PR would be a lot easy if we had a `SaveButton` component as we would change only one file. I suggest creating a new PR for that. Maybe a generic component for all types of buttons, Save, View, Create..

src/styles/main.scss
codegaze: Updated. My monitor home shows colors by proximity. :joy:

mertkahyaoglu:
Thanks @codegaze 🎉

PR #188, opened 2016-08-25
Comment out hints in Issue Template

ashmaroli:
Since the hints are only really required while drafting the issue, lets comment them out of otherwise render.


benbalter:
Thanks @ashmaroli!

ashmaroli:
You're welcome. Thanks for the merge. :smiley:

PR #189, opened 2016-08-25
Update front-end to a sleek interface

ashmaroli:
Proposal to change admin front-end to a sleek interface.

jekyll-admin-style

(links in action-rows changed to a button-appearance that get colored on hover.)


mertkahyaoglu:
Thanks @ashmaroli . Really loved the action buttons ❤️ 

I just have some suggestions regarding the interface.

* I would prefer some spaces between sidebar links. Looks like they stacked on the top.

* I would prefer a different color when hovered on the sidebar links like the current one.

* The black `Splitter` in editing view doesn't look good. I think the current one's color is a little bit more charming 😬  The ones in the sidebar are perfect though.

screen shot 2016-08-25 at 6 27 58 pm

* You have to use `className` instead of `class`. Therefore those classes are not applied to action buttons. Here how it looks like now;

screen shot 2016-08-25 at 6 28 09 pm

* There should be some padding after the icon. It would be great if you can put it back. Otherwise, it looks like below 😬 

screen shot 2016-08-25 at 7 01 01 pm

screen shot 2016-08-25 at 6 27 58 pm

After those suggestions there is not much left 😬  but again I really like action buttons and the color of the sidebar links. For me, these would be enough for this PR.

@benbalter What do you think?

ashmaroli:
Thank you for the feedback @mertkahyaoglu, I must confess I did not test the changes I proposed, hence could not see those bugs you mentioned.

> I would prefer a different color when hovered on the sidebar links like the current one.

doesnt it already? 
from pale-skin-tone to orange?

DirtyF:
IMHO rather than adding new styles, we should aim at cohesiveness. 

Size, colors and icons differ depending on screens.

jekyll-admin-buttons

A basic styleguide could help to keep on overview on the avalaible styles

Please note that Kirby gets the job done without having to rely on colors to convey information.

Grav has done a pretty good job at providing a colorful experience and keeping a cohesive UI.
https://getgrav.org/blog/beta-admin-plugin-available

It would be awesome to have a true designer's help on this, it's a different job from coding a React app or writing CSS.

ashmaroli:
@DirtyF, I did see the need for cohesiveness too. Thought of doing that in another PR though..

ashmaroli:
updated to fix bugs and consider cohesiveness.
(tested on a windows machine)

src/styles/main.scss
```diff
@@ -67,33 +70,40 @@ i {
...
+  }
 }
 
 .btn-active {
   @include btn($btn-active, $btn-active-border);
 }
 
 .btn-inactive {
-  @include btn($btn-inactive, $btn-inactive-border, default, .5, false); 
+  @include btn($btn-inactive, $btn-inactive-border, default !important, .5 !important, false);
```
ashmaroli: no change in cursor or opacity on hover, if `.btn-inactive`

src/styles/main.scss
```diff
@@ -67,33 +70,40 @@ i {
...
+  @include btn($btn-inactive, $btn-inactive-border, default !important, .5 !important, false); 
 }
 
 .btn-success {
-  @include btn($btn-success, $btn-success-border);
+  @include btn($btn-success, $btn-success-border, pointer);
+  &:hover {
+    background-color: lighten($btn-success, 5%) !important;
+    @include box-shadow(0 1px 0 $btn-success-border);
+  }
```
ashmaroli: added to override orange color when hovering on `.btn-fat`

src/styles/mixins.scss
```diff
@@ -48,18 +48,18 @@
   transform: translateX($px);
 }
 
-@mixin btn($color, $border-color, $cursor: pointer, $opacity: 1, $hover: true) {
-  cursor: $cursor!important;
+@mixin btn($color, $border-color, $cursor: pointer, $opacity: 1, $hover-fx: true) {
+  cursor: $cursor;
```
ashmaroli: removed `!important` to allow btn-inactive to maintain non-hover style on _mouseover_.
/cc @codegaze

src/styles/variables.scss
```diff
@@ -3,18 +3,24 @@ $content-side-width: 260px;
...
 $btn-initial-border: #828282;
+$btn-fat-hover: $dark-orange;
+$btn-fat-hover-border: $dark-orange;
 $btn-active: #ff6e00;
 $btn-active-border: #dd5c00;
-$btn-inactive: $btn-initial;
-$btn-inactive-border: $btn-initial-border;
+$btn-inactive: $btn-initial !important;
+$btn-inactive-border: $btn-initial-border !important;
 $btn-success: #8dcf3f;
```
ashmaroli: To make sure color doesn't change even if another attribute is added to parent element

mertkahyaoglu:
Great work!

LGTM except this;

screen shot 2016-08-28 at 1 07 57 pm

@ashmaroli Shall we align them by removing [this line](https://github.com/ashmaroli/jekyll-admin/blob/dd97b95b8a9fbe45c8c7b31cd922b1f0ba80a422/src/styles/content.scss#L59) ?

ashmaroli:
Nice catch! :smiley: 
Instead of removing that line, and inheriting `vertical-align: middle` from tbody, its better to have it set to `middle` incase tbody styles change for some reason.
I'll push the update right away.

DirtyF:
Good progress towards cohesiveness, well done @ashmaroli, looks already more polished.

I have a couple of questions : 

1. Why three types of grey for inactive buttons ?

buttons-grey

Moreover the lighter grey doesn't provide a good contrast, it harms accessibility IMHO. I would try a dark grey on these, or do at least some experiments and iterations.

2. Why is the meaning of the colors?  Why the `View` button does not have the same active color on the list page and on the edition page?

page-buttons

3. Could we improve cohesiveness when a field get the focus ?

The current behavior on search is good.

focus-search

Could we do the same when a user edit a title or a filename ? Should we just change the color of the cursor to show that this area is active ?

focus-finename
focus-title

focus-editor

Border colors do not seem to inherit same colors, thickness. Alignement could be improved I guess.

4. Why a different input for the key and value for YFM ?

edit-yfm-key
edit-yfm-value

Hope this helps, small details matter.

ashmaroli:
Hi @DirtyF, lot of thanks for the critique.
> small details matter

**I couldn't agree more. esp., when the stake is on cohesiveness**. :smile: 

To answer your queries:
- its not 3 different grey buttons there. but 3 `$grey` buttons with one of them at 0.5 opacity (from a recently merged PR). The light grey is intended to mimic a disabled state. Accessibility doesn't matter (much) as its supposed to be a disabled button. When the fields are changed, the button turns green and stays green to denote the new active state. The view button here takes on the `$primary-color` from the site's palette, when `onmouseover`.
- _change the color of the cursor_, (is that possible?). But yes, I can denote focus-state by having a colored bottom-border `onfocus`
- Agree with you, totally, on the fields-onfocus-discrepancies. It was just lapse on my side.
- Should the 'Delete' button be there? I'll leave that decision to the maintainers.

Keep the feedback coming. Happy to incorporate changes.
Disclaimer: the variables mentioned above are just examples and not really defined in our sass files.

mertkahyaoglu:
> change the color of the cursor_, (is that possible?)

Maybe `cursor: no-drop` ?

README.md
```diff
@@ -4,7 +4,7 @@
 
 A Jekyll plugin that provides users with a traditional CMS-style graphical interface to author content and administer Jekyll sites. The project is divided into two parts. A Ruby-based HTTP API that handles Jekyll and filesystem operations, and a Javascript-based front end, built on that API.
 
-![screenshot of Jekyll Admin](https://cloud.githubusercontent.com/assets/282759/17258537/62e23ed6-5595-11e6-89b0-31c787f0492a.png)
+![screenshot of Jekyll Admin](/spec/screenshot.png)
```
mertkahyaoglu: Let's not add it to `spec`. @benbalter might want to add it in a separate PR which loads the updated screenshot from the cloud.

ashmaroli: okay, should I wait for his input? or reset back to prev commit right now?

mertkahyaoglu: We are using `spec` folder for test purposes only. I don't think the screenshot belongs there. Let's wait for his approval to make sure.

ashmaroli: @benbalter, should I remove the latest commit or move the screenshot to the root?

mertkahyaoglu:
Thanks @ashmaroli 👍

PR #192, opened 2016-08-26
Update package to allow running script on windows

ashmaroli:
I had trouble running `script/server-frontend` as `script/test-server` call further ahead would not run out-of-the-box on Windows.
```
'script' is not recognized as an internal or external command,
operable program or batch file.
```
Hence prepended a `sh`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "start-message": "babel-node tools/startMessage.js",
     "prestart": "npm-run-all --parallel start-message remove-dist",
     "start": "npm-run-all --parallel start-api test:watch open:src lint:watch",
-    "start-api": "script/test-server",
+    "start-api": "sh script/test-server",
     "open:src": "babel-node tools/srcServer.js",
     "lint": "esw webpack.config.* src tools",
     "lint:watch": "npm run lint -- --watch",


benbalter:
Since https://github.com/jekyll/jekyll-admin/blob/master/script/test-server#L1 declares it as shell, should it be `sh` and not `bash` that's prepended?

ashmaroli:
it works with either `sh` or `bash` I'm not very familiar with the exact differences between the two. So, its your call. i'll update the PR accordingly.

benbalter:
I don't believe we're using any bash-isms now, but it's likely best to use the same interpreter in both environments, so for now, if it works on windows, I think `sh` is best.

ashmaroli:
would this change have bumped the minor-version instead of the patch-version?

mertkahyaoglu:
I think it's better to fix all of the Windows issues in a single patch.

In addition, I haven't run a complete test on a Windows machine 😬

ashmaroli:
Okay. Any plans to have Appveyor detect those bugs automatically?

benbalter:
> would this change have bumped the minor-version instead of the patch-version? 

I believe this can be a patch fix. @ashmaroli thanks for this! If there are any other known Windows issues, please feel free to open issues, or even better, submit PRs with tests, since we now have appveyor running on each pull request.

PR #194, opened 2016-08-26
New Issue Template

ashmaroli:
Drafted a new template for submitting Issues.
- Commented out any existing hints.
- Added placeholder texts for certain sections.
- Added descriptive sub-headings.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,25 +1,25 @@
-Describe in one or two sentences why you're opening this issue:
+### Description:
+<!-- Describe in one or two sentences why you're opening this issue: -->
+> **REPLACE-ME**
+
+### Tell us a bit about yourself:
+* Version of JekyllAdmin I'm using <HINT: use `bundle show` to check>: ***REPLACE-ME***
+* Version of Jekyll I'm using <HINT: use `bundle show` to check>: ***REPLACE-ME***
+* Operating System <e.g. OS X, Windows>: ***REPLACE-ME***
+* Browser <e.g, Safari, Chrome>: ***REPLACE-ME***
+
+### Steps to reproduce:
+<!-- List the steps someone else can take to reproduce the conditions you're reporting. -->
 
 
-### Tell us a bit about yourself
+### I expected the following:
+<!-- After you take those steps, what do you expect to happen? -->
 
-* Version of JekyllAdmin I'm using <hint: use `bundle show` to check>:
-* Version of Jekyll I'm using <hint: use `bundle show` to check>:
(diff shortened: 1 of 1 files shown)


ashmaroli:
I've opened a new issue ( No. 197 ) with this template to additionally serve as a live preview for this PR.

ashmaroli:
@mertkahyaoglu, I think `documentation` is actually an incorrect label here. This PR doesn't affect JekyllAdmin documentation, only the Issue drafting here on GitHub.  
A better alternative would be `GitHub Repo` or similar.  :smiley:

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,25 +1,25 @@
-Describe in one or two sentences why you're opening this issue:
+### Description:
+<!-- Describe in one or two sentences why you're opening this issue: -->
+> **REPLACE-ME**
```
benbalter: Why should the issue description be a block quote? What are they quoting?

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,25 +1,25 @@
...
 
+### Tell us a bit about yourself:
+* Version of JekyllAdmin I'm using <HINT: use `bundle show` to check>: ***REPLACE-ME***
```
benbalter: Why add the `**REPLACE ME**` text? I feel like that's creating more work for a bug reporter (highlight and delete 5 times), than just having a colon and a blank which should achieve the same effect.

benbalter:
Thanks @ashmaroli! It may also be worth noting, either in this PR or a subsequent follow up, that you should report *one* bug per issue, not multiple, where they can be distinguished.

.github/ISSUE_TEMPLATE.md
ashmaroli: :smile: themselves? lol!
I placed it there for the formatting. But I now realize why that's wrong. So, I'll remove the blockquote formatting.

.github/ISSUE_TEMPLATE.md
ashmaroli: I wanted the issue drafter to insert the info exactly between the three stars to have their response formatted. Rarely does a user style their issue draft.
I'll remove those too..

ashmaroli:
uhm.. this is one, single issue.
I understand what you mean though.

benbalter:
> It may also be worth noting, either in this PR or a subsequent follow up, that you should report one bug per issue, not multiple, where they can be distinguished.

Don't have to do it here, I was suggesting that we add a note with that reminder to the form (not that you were breaking the rule!) :smile:

PR #199, opened 2016-08-27
Initial commit for a Button abstraction

codegaze:
This is a first draft of a button component. Has three predefined versions/types 'create', 'save' & 'delete' and also custom buttons with the option of adding classes and text.

I changed the element from link to buttons(with disabled value) in favor of better semantics, There is no 'view' predefined button there because it makes more sense to be left as a link(I think).

For now I just wanted to know if you think this is the right approach. 
I'm not really excited about the predefined buttons 'config' and I'm not really sure about the testing.
Should this test all the predefined versions with true false values(like the 'create' button)? Should I take another approach? 

If you think this could be improved and merged, I'm open to suggestions :beers:

### DIFF
--- a/src/components/form/Button.js
+++ b/src/components/form/Button.js
@@ -1,0 +1,55 @@
+import React, { Component, PropTypes } from 'react';
+
+export default class Button extends Component {
+
+
+  handleClick(e){
+    const { onClick } = this.props;
+    onClick(e);
+  }
+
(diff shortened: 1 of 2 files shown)


mertkahyaoglu:
Closing this after #189. @codegaze If you'd like to continue to this later, we can work on a new PR.

PR #200, opened 2016-08-28
[wip] /admin should redirect to posts by default

### LINKED ISSUE #156: Dashboard main page, rather than being welcomed by posts or pages
Right now when I enter Jekyll admin, I'm greeted my `pages` directory, which I'm rarely going to edit. Though this behavior might be specific to me since I included a `_pages` directory. But either way, whether posts or pages, it would be more pleasing to be greeted by a dashboard main page.

DirtyF:
`/admin` should redirect to `/admin/collections/post` instead rather than `admin/pages`

fix #156


DirtyF:
@mertkahyaoglu I can't figure out where the main redirection happens, I thought changing `index.js` would be sufficient but it does not seem to be the case.

ashmaroli:
Try playing around with [this](https://github.com/jekyll/jekyll-admin/blob/master/src/routes.js#L21)

mertkahyaoglu:
> Try playing around with [this](https://github.com/jekyll/jekyll-admin/blob/master/src/routes.js#L21)

Yep, that's it.

mertkahyaoglu:
@DirtyF Can you update the branch? I think we can merge this now. Don't know why we are keeping this 😬

DirtyF:
@mertkahyaoglu Ideally a dashboard page would be much convenient to access latest edited documents ou create new ones but for now, it seemed like a reasonable default. Didn't play enough with `react-router` to modify current routes.

ashmaroli:
@mertkahyaoglu, can we have the collections-items fold into a list-item labelled *Collections* instead of listing all collections in the sidebar by default? I'll open a dedicated ticket, if this holds merit..

mertkahyaoglu:
@DirtyF 👍  for dashboard page but we need the backend support for that.

@ashmaroli I'm in favor of instant access. As an end user, I don't want to click one extra link to go to a specific collection. Maybe we can add an extra splitter to separate them visually.

ashmaroli:
> As an end user, I don't want to click one extra link to go to a specific collection.

Edge-Case: A site with **lots** of collections will push the remaining 3 important list-items down that would force the user to scroll the sidebar vertically to access `Data Files`, `Static Files` and `Configuration`

mertkahyaoglu:
I'm closing this. Let's continue from #156

DirtyF:
@mertkahyaoglu Sorry for not getting back to you sooner on this. What is missing here?

mertkahyaoglu:
Closed this in favor of a dashboard page. But if you think it would be good to have `posts` at the top since then, let's reopen this.

Sorry if I'm being rude for closing this, I thought we decided on the dashboard page

PR #212, opened 2016-08-31
gitignore .DS_Store

mertkahyaoglu:
@benbalter How did we miss this? I suddenly started to see `.DS_Store` files 👀


benbalter:
Ha. Good call.

PR #217, opened 2016-09-02
Add version info to sidebar

ashmaroli:
This PR adds version info to the sidebar.
It doesn't solve any existing issues but simply acts as filler content at the bottom portion of the sidebar.
_To me,_ this adds a sort of visual balance.

version-on-sidebar

### DIFF
--- a/src/containers/Sidebar.js
+++ b/src/containers/Sidebar.js
@@ -3,6 +3,7 @@ import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
 import { Link } from 'react-router';
 import _ from 'underscore';
+import { VERSION } from '../utils/version';
 
 // Constants
 import { ADMIN_PREFIX } from '../constants';
@@ -68,6 +69,10 @@ export class Sidebar extends Component {
               <Link activeClassName="active" to={`${ADMIN_PREFIX}/configuration`}><i className="fa fa-cog"></i>Configuration</Link>
             </li>
           </ul>
+        </div>
+        <div className="about_info">
+          <p className="module_name">JekyllAdmin</p>
+          <p className="module_version">VERSION: {VERSION}</p>
         </div>
       </div>
     );
--- a/src/utils/version.js
+++ b/src/utils/version.js
@@ -1,0 +1,1 @@
+export const VERSION = '0.1.1';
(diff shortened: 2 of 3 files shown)


src/utils/version.js
```diff
@@ -0,0 +1 @@
+export const VERSION = '0.1.1';
```
ashmaroli: Is there a way to get this value from the `version.rb`

ashmaroli:
Any particular reason Appveyor check did not run on this branch?

benbalter:
@mertkahyaoglu what do you think?

src/containers/Sidebar.js
```diff
@@ -69,6 +70,10 @@ export class Sidebar extends Component {
             </li>
           </ul>
         </div>
+        <div className="about_info">
+          <p className="module_name">JekyllAdmin</p>
+          <p className="module_version">VERSION: {`${VERSION}`}</p>
```
mertkahyaoglu: Just use `{VERSION}` here.

mertkahyaoglu:
Really catching my 👀 . I suggest making it smaller because it is not a priority thing to see.

src/utils/version.js
ashmaroli: @mertkahyaoglu  any suggestion for :point_up_2: ?

ashmaroli:
> I suggest making it smaller 

Done.

src/containers/Sidebar.js
ashmaroli: updated.

src/utils/version.js
mertkahyaoglu: It could be possible if `/_api/` endpoint provides such information but currently no 😕

mertkahyaoglu:
@ashmaroli Can we fix this gap?

screen shot 2016-09-06 at 8 45 08 pm

ashmaroli:
--updated--

mertkahyaoglu:
What if I have too many collections ?

screen shot 2016-09-06 at 11 22 50 pm

ashmaroli:
@mertkahyaoglu, The optimal solution would be to have the `div.routes` at a fixed-height and `overflow-y: auto` 
For a PR that's purely cosmetic , the resulting scrollbar (since scrollbars can be styled only for webkit browsers) is a tacky entity.
So, closing this..

PR #224, opened 2016-09-21
Autogenerate filename from title if not given

### LINKED ISSUE #105: Create new page : placeholder and valid user input
Some UX feedback after testing the creation of a new page: 

* The placeholder when creating a new page is _Filename_, where as for a new post it is a valid example. I would suggest also a valid example here.

* There does not seem to be a validation on what is entered by the user when creating a page (or a post). So I can type "A new file" with spaces and no extension. But if I try to preview that kind of page, it will download the file instead of displaying it.

mertkahyaoglu:
With this PR, filenames are auto-generated by slugifying titles if users does not provide them for pages and documents. If they do, validations are applied as before.

@benbalter , @DirtyF What do you think?

Closes #105


DirtyF:
@mertkahyaoglu Nice addiction. I ran some quick tests, it seems to work as expected.

Juste one question : Could we enhance the slugify helper in order to add a separator for ' and ‘ chars? Often used in french, e.g. Démonstration par l'exemple becomes demonstration-par-l**-**exemple.md

But maybe it's something you might want to add later with i18n and is out of the scope of this PR.
Anyway, it's already a time saver and a nice feature for users. 👌

mertkahyaoglu:
@DirtyF I think [node-slug](https://github.com/dodo/node-slug) will do the job. it's respecting [RFC 3986](https://tools.ietf.org/html/rfc3986).

I don't know which one is correct but according to `node-slug`, your example results in `demonstration-par-lexemple` which is one dash short at the end.

DirtyF:
@mertkahyaoglu I can see that node-slug tests remove this kind of chars, I didn't know about the RFC. Let's go with node-slug implementation then.

ashmaroli:
There seems to be a bug related to this PR.. 
When I try to run `script/server-frontend` aginst the `master` branch, I get the follwing error:

ERROR in ./src/utils/helpers.js
Module not found: Error: Cannot resolve module 'slug' in C:\~\src\utils
 @ ./src/utils/helpers.js 16:12-27

mertkahyaoglu:
@ashmaroli Did you `npm install` after merging with master? This PR adds a new npm dependency `slug`.

ashmaroli:
yay! its working again.. :smiley: 
I was not aware of the need to run `npm install`. All is well now.

PR #226, opened 2016-09-28
Use original SimpleMDE

mertkahyaoglu:
Jekyll Admin was using the fork of [SimpleMDE](simplemde.com) due to some problems with spell checking feature. This PR resolves the issue with the help of Webpack and uses the original repo back again.


ashmaroli:
The build error mentioned in the PR body was due to a lapse on my part.

mertkahyaoglu:
Ops. Yeap, that was for `codemirror` error 👍 .

PR #227, opened 2016-10-06
update arguments to `server_address`

Crunch09:
see jekyll/jekyll#5451

### DIFF
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -12,12 +12,7 @@ module Jekyll
 
           jekyll_admin_monkey_patch(server)
 
-          Jekyll.logger.info "Server address:", server_address(
-            server.config[:SSLEnable],
-            server.config[:BindAddress],
-            server.config[:Port],
-            opts["baseurl"]
-          )
+          Jekyll.logger.info "Server address:", build_server_address(server, opts)
           launch_browser server, opts if opts["open_url"]
           boot_or_detach server, opts
         end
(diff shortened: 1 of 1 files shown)


Crunch09:
please see discussion in jekyll/jekyll#5451 , maybe it would be wiser to revert the change which caused this issue because it was only a refactoring.

benbalter:
So the issue is that we were using a private API, and that API was refactored. Instead of monkey patching `process`, why don't we monkey patch the refactored method, `start_up_webrick`, meaning we'd just need to move `jekyll_admin_monkey_patch(server)` immediately after ` server.mount(opts["baseurl"], Servlet, destination, file_handler_opts)`.

If we really wanted, we could condition the old and new monkey patch based on the Jekyll version, but given how young the project is, I'm fine requiring Jekyll >= 3.3.

/cc @parkr in case you feel otherwise.

parkr:
I'm going to revert the signature change.

parkr:
> If we really wanted, we could condition the old and new monkey patch based on the Jekyll version, but given how young the project is, I'm fine requiring Jekyll >= 3.3.

I see now. I'm 👍 on this kind of approach. Can this be released today? I'd rather wait until this patch is released to release Jekyll 3.3.

benbalter:
> Can this be released today?

Chicken and egg. The cleanest way to do the new monkeypatch would be to require >= 3.3. I think it's fine if Jekyll Admin doesn't work on Jekyll 3.3 for a day or two, and we can require 3.3 going forward.

Crunch09:
@benbalter Should i update this PR with the changes you suggested or do you work on a new PR?

benbalter:
@Crunch09 if you want to put together a PR over to JekyllAdmin, I'd be happy to review it and get it merged. 👍

Crunch09:
As the `server_address` change was reverted, closing this in favor of #228 .

PR #228, opened 2016-10-07
monkeypatch `start_up_webrick` instead of `process`

Crunch09:
We need to require jekyll ~> 3.3 for `start_up_webrick` to be present. For more background see #227 .

I tried to write a spec for it but i wasn't able to stop webrick after the test run without also killing the test run itself 😕 

cc: @benbalter


benbalter:
@Crunch09 thanks!

> I tried to write a spec for it but i wasn't able to stop webrick after the test run without also killing the test run itself

No worries. Wrote a quick integration test via https://github.com/jekyll/jekyll-admin/commit/2f64d3447fd907b91440d7d9be951509598d732f.

Crunch09:
@benbalter Great, thank you!

mertkahyaoglu:
> No worries. Wrote a quick integration test via 2f64d34.

@benbalter this test is failing on travis

```
Failure/Error: expect(response.code).to eql("200")
  expected: "200"
  got: "404"
```

How can we fix this?

PR #230, opened 2016-10-12
Add yarn.lock file

mertkahyaoglu:


benbalter:
@mertkahyaoglu can you explain a bit more about what this does?

mertkahyaoglu:
[Yarn](https://github.com/yarnpkg/yarn) is a new package management tool like `npm` (also compatible with it) and it is really fast. This PR adds the [lock file](https://yarnpkg.com/en/docs/yarn-lock) to provide faster module downloads needed for the frontend.

PR #233, opened 2016-10-26
Add toolbar save icon for markdown editor

### LINKED ISSUE #157: Ability to save in fullscreen editor
While using the fullscreen editor (which looks great btw), it would be helpful to have a save button on the right side of the toolbar. You can't really use cmd+s on the browser to save changes, and it would be annoying to have to exit fullscreen every time you wanted to save. Plus, the right side of the toolbar is empty, so why not (:

mertkahyaoglu:
Closes #157

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,3 +6,6 @@ AllCops:
     - lib/jekyll-admin/public/**/*
     - src/**/*
     - node_modules/**/*
+
+Metrics/BlockLength:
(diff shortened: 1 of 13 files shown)


mertkahyaoglu:
Wow. We should definitely fix these checks 😭

ashmaroli:
the ruby failure seems to be due to recently modified monkey-patch..

PR #234, opened 2016-10-29
Metafields design tweaks

### LINKED ISSUE #139: Creating a New Metadata Field Doesn't Automatically Create A Valid Field Name, Or Indicate the User Can
When a user adds a "New meta-data field" they should 
1) not have to rename the field by default, and 
2) it should be intuitive that there's a way to do so, and
3) the field should instruct them how to make a valid field name, or not allow them to make an invalid field.

The current view (screenshot attached) creates an invalid field name by default (e.g. `New Field 14`), and there's no way of telling that the user can click into the box to edit it, nor is there any way to enforce or instruct them to use a proper field name.

screen shot 2016-08-18 at 7 44 48 pm

mertkahyaoglu:
Closes #139 

screen shot 2016-10-29 at 6 19 48 pm

### DIFF
--- a/src/components/metadata/MetaButtons.js
+++ b/src/components/metadata/MetaButtons.js
@@ -25,7 +25,7 @@ export class MetaButtons extends Component {
             onBlur={() => onDropdownBlur()}
             className="meta-button"
             tabIndex="-1">
-              <i className="fa fa-ellipsis-h"></i>
+              <i className="fa fa-chevron-down"></i>
           </button>
           <div className="dropdown-wrap">
             {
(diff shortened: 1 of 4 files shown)


ashmaroli:
since the key-field class has a max-width, can we restrict and inform users to keep the key.length within a certain no. of characters?
I suspect a *super-long* key will have characters towards the end hidden..
ja

mertkahyaoglu:
That's a good idea. Could you create a follow-up PR for that?

ashmaroli:
okay.. but I wont be able to add the validation alert..

mertkahyaoglu:
What about adding just `maxlength` attribute? Would that be enough?

ashmaroli:
that would be enough.. will create a PR soon..

PR #237, opened 2016-10-30
fix z-index bug in classname meta-buttons

ashmaroli:
the meta-buttons are currently visible even when the full-screen editor is active.
This patch fixes that by reducing the `z-index` attrib.


mertkahyaoglu:
Nice catch 👍

PR #238, opened 2016-10-30
restrict meta-key length

ashmaroli:
restrict metadata-key length to a max 21 characters in compliance with current style definitions
ja2

### DIFF
--- a/src/components/metadata/MetaField.js
+++ b/src/components/metadata/MetaField.js
@@ -53,7 +53,8 @@ export class MetaField extends Component {
             defaultValue={fieldKey}
             className="field key-field"
             type="text"
-            placeholder="Key" />
+            placeholder="Key"
+            maxLength="21" />
           <MetaButtons
             currentType={type}
             parentType="top"


benbalter:
> in compliance with current style definitions

What do you mean by that?

ashmaroli:
> in compliance with current style definitions
>
> What do you mean by that?

`.meta-key .key-field` class currently has a `max-width` of `190px` which supports just 21 characters to be visible. If at a later date, the max-width changes, the `maxlength` attrib can be changed accordingly.

mertkahyaoglu:
@benbalter What do you think about this? Shall we limit meta key length? Another option would be to increase input width.

ashmaroli:
@DirtyF, why the downvote?

benbalter:
The text input may be limited to 21 characters visible at any given time, but there's no reason meta keys can't extend well beyond that. We shouldn't be limiting our data model due to limitations in the UI. If anything, the opposite should be the case. What's the harm with the text scrolling within the textbox for users with large keys?

ashmaroli:
point taken.

PR #240, opened 2016-10-30
Confirm navigation if there are unsaved changes

### LINKED ISSUE #158: Prompt user to confirm before leaving page with unsaved changes
Right now if you have any unsaved changes and click on a different page, your progress isn't saved and you aren't prompted to confirm the decision. It would be safer if you were prompted to confirm whether you want to stay or leave before leaving any page with unsaved changes. It'd suck to lose all your progress because you accidently hit the back button.

mertkahyaoglu:
Fixes #158

### DIFF
--- a/src/containers/views/Configuration.js
+++ b/src/containers/views/Configuration.js
@@ -1,6 +1,7 @@
 import React, { Component, PropTypes } from 'react';
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
+import { withRouter } from 'react-router';
 
 // Components
 import Editor from '../../components/Editor';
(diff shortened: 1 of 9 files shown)


src/containers/views/Configuration.js
```diff
@@ -10,6 +11,17 @@ import { putConfig, onEditorChange } from '../../actions/config';
...
 
+  componentDidMount() {
+    const { router, route } = this.props;
+    router.setRouteLeaveHook(route, this.routerWillLeave.bind(this));
+  }
+
+  routerWillLeave(nextLocation) {
+    const { editorChanged } = this.props;
+    if (editorChanged)
+      return 'Your work is not saved! Are you sure you want to leave?';
```
ashmaroli: somehow `Your work is not saved!` doesn't sound (grammatically) correct to me.. may i suggest:
**`Your work has not been saved!`** ? or **`You've unsaved changes on this page.`** ?

PR #248, opened 2016-12-16
Separate frontend messages into a file

### LINKED ISSUE #241: Separate frontend messages into a file
Currently some messages like the followings;

* `Are you sure that you want to delete FILENAME ?`
* `No data files found.`
* `Are you sure that you want to delete FILENAME ?`

are hardcoded into components (e.g. [here](https://github.com/jekyll/jekyll-admin/blob/master/src/containers/views/DocumentEdit.js#L61)). When we want to change the `delete message`, for example, we have to make changes to all of files that contains it. It would be nice to separate all messages into a file like `messages.js` so we can make changes to only that file when we want.

In addition, this would help us implement #99 easily in the future.

mertkahyaoglu:
Resolves #241

### DIFF
--- a/docs/frontend/constants.md
+++ b/docs/frontend/constants.md
@@ -11,10 +11,12 @@ Stores all of the action creators' names
 
 Stores the following messages;
 
-* delete message
-* leave message
-* override static file message
+* confirm delete message
+* confirm leave message
+* confirm override static file message
 * not found message
+* notification messages
+* validation messages
 
 This allows us to control all of the messages in a single place and also will help us localize the admin panel in the future.
 
--- a/docs/frontend/constants.md
+++ b/docs/frontend/constants.md
@@ -7,6 +7,17 @@ permalink: /frontend/constants/
 
 Stores all of the action creators' names
 
+## Messages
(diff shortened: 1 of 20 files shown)


ashmaroli:
I think there should also be a set of sort of success messages:
```
"[FILE] deleted."
"[FILE] created."
"[FILE] renamed."
"Changes saved successfully"
```

mertkahyaoglu:
Totally forgot them. Thanks 👍

mertkahyaoglu:
I don't know if there is a better way for handling these messages. I'm open to suggestions to improve the readability.

ashmaroli:
@mertkahyaoglu I'm running the `server-frontend` script on latest `master` and:
- creating a new `Page` / `Post` does not display any flash messages.
- Deleting an existing `Page` / `Post` does not display a successfully deleted flash message.
- Deleting an uploaded static file does not display a flash message.

mertkahyaoglu:
Yeap, those messages are intentionally left missing because after those actions you can see the results directly. But for completeness, we should still add them.

Issue #246, opened 2016-12-02
Editor breaks configuration file and datafiles


ashmaroli:
### Description:

The Editor under /configuration and /datafiles breaks datafiles containing the fold character and comments`>`

### Tell us a bit about yourself:
* Version of JekyllAdmin I'm using <HINT: use `bundle show` to check>: 0.1.1
* Version of Jekyll I'm using <HINT: use `bundle show` to check>:3.3.1
* Operating System <e.g. OS X, Windows>:**_Windows 7_**
* Browser <e.g, Safari, Chrome>:Chrome

### Steps to reproduce:

- Copy `email:` and `description:` from a Jekyll site's `_config.yml`
- Add a few lines of comments to the **spec's config file**
- Run `script/server-frontend`
- Edit the **spec's config file** by pasting the content copied earlier.
- Hit Save
- Examine the **spec's `_config.yml`**
### I expected the following:

- The file content to be a valid YAML file
- The comments to not be deleted.

### But got the following, instead:

- The file to be broken after `description:`
- Existing comments deleted

### Other details:

N/A

mertkahyaoglu:
@benbalter @parkr Is there a way the api can send the raw configuration as yaml rather than [json](https://github.com/jekyll/jekyll-admin/blob/master/lib/jekyll-admin/server/configuration.rb#L5)? This way, the front end does not have to parse it back to YAML, which will preserve comments.

parkr:
That's a great idea! You would have to re-read the file but yeah we should be able to read the file as-is and allow it to be written again as-is. @benbalter: sound possible to you?

ashmaroli:
fixed.

mertkahyaoglu:
@ashmaroli How is this fixed ? 😄

ashmaroli:
@mertkahyaoglu this issue is no longer reproducible on `master` :smiley:

mertkahyaoglu:
Awesome then :man_dancing:

PR #255, opened 2016-12-23
Use `raw_content` for PUT requests to support other formats

### LINKED ISSUE #106: Ability to create/update JSON data files

### LINKED ISSUE #108: Ability to read/write CSV data files
Similar to https://github.com/jekyll/jekyll-admin/issues/106, this would allow the API to read, write and delete CSV data files in the `_data` folder, which Jekyll supports.

mertkahyaoglu:
With this PR, the front end sends PUT requests for data files with `raw_content` instead of `content`. This way, the backend saves the content of data files to disk as they are rather than parsing to `yaml` ([code](https://github.com/jekyll/jekyll-admin/blob/master/lib/jekyll-admin/server/data.rb#L37)). This will allows us to create/update data files in any format.

Closes #106 and closes #108.

### DIFF
--- a/src/containers/views/Configuration.js
+++ b/src/containers/views/Configuration.js
@@ -5,6 +5,7 @@ import { withRouter } from 'react-router';
 import Editor from '../../components/Editor';
 import { putConfig, onEditorChange } from '../../actions/config';
 import { getLeaveMessage } from '../../constants/messages';
+import { toYAML } from '../../utils/helpers';
 
 export class Configuration extends Component {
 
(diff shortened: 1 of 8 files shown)


src/containers/views/Configuration.js
```diff
@@ -43,7 +44,7 @@ export class Configuration extends Component {
         <Editor
           editorChanged={editorChanged}
           onEditorChange={onEditorChange}
-          json={config}
+          content={toYAML(config)}
```
mertkahyaoglu: [reference](https://github.com/jekyll/jekyll-admin/issues/246#issuecomment-268996338) to another issue

PR #267, opened 2017-01-09
Use 'monokai' theme in `Editor` component

ashmaroli:
The current theme `github` used in the [**`Editor.js`**](https://github.com/jekyll/jekyll-admin/blob/master/src/components/Editor.js#L30) component is rather bland and doesn't seem to impart optimal syntax highlighting.

This PR proposes to use the `monokai` theme.
- optimal syntax-highlighting
- dark-background offers optimal contrast.

theme


mertkahyaoglu:
Thanks @ashmaroli 👍

PR #268, opened 2017-01-10
Remove unused API fields

mertkahyaoglu:
Some of the fields were not being used by the front end. This PR removes them from the API output.

These are the omitted fields;

- `content` => front end uses `raw_content` instead in order to keep the files as they are
- `excerpt`
- `output` 
- `next` & `previous` documents
- `url` => we have `http_url`

### DIFF
--- a/lib/jekyll-admin/apiable.rb
+++ b/lib/jekyll-admin/apiable.rb
@@ -3,8 +3,13 @@ module JekyllAdmin
   # additional, API-specific functionality without duplicating logic
   module APIable
 
-    CONTENT_FIELDS = %w(next previous content excerpt).freeze
-
+    # - Documents have duplicate output and content fields, Pages do not
+    # Since it's an API, use `raw_content` in both for consistency
+    # - API doesn't need `content` & `excerpt` since we have `raw_content`
+    # which is the only field that the front end deals with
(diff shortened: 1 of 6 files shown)


benbalter:
I'd just caution, that there's little downside to adding an extra field to an API response, especially locally, as in the worst case scenario, it's already generated, there's no bandwidth cost, and it's just ignored. 

At the same time, I'd push us to think about the API we're putting out (like all APIs) as a general Jekyll site API, rather than an API built for one specific client. While this client may not need those fields, it's not hard to imagine user workflows that would.

For example, what if we separated the plugin into two gems, Jekyll-API and Jekyll-Admin in which Jekyll-API put out an API (maybe as read only static JSON, maybe as an authenticated RESTful service) and Jekyll-Admin simply package the front end. Would this be the right decision for Jekyll-API then?

Especially with a single local client, it's easy enough to make breaking changes to the API, but I'd rather think about this as building a general Jekyll site API / interaction model, and a the first of many clients, rather than simply (and perhaps unnecessarily) optimizing how to local services talk to one another.

mertkahyaoglu:
@benbalter That totally makes sense 👍  I wasn't thinking the big picture. Thank you for the explanation master!

PR #272, opened 2017-01-13
Mimic Metafields placement on actual documents

ashmaroli:
- Move Metafields above the document body to mimic placement of Front Matter.
- initialize a metafield pair for `layouts` by default as every Jekyll Document has a `layout` metadata at minimum.

preview:

metafields


mertkahyaoglu:
- I don't think it's a good idea to move `metafields` to the top. They might grow too much and fill the whole screen. The content is the main focus for the most part and should be at the top to be editable easily. You don't change front matters so often.

- I thought `document` is a general term for all types of collection items. If we say `New Post`, do we cover other collections, not just posts, as well? Because that button is reused by other collections as well.

DirtyF:
FYI, here is how some similar products display metafields: 

- Siteleaf place metafiels **after** the content (like Jekyll-Admin)
- CloudCannon allow to hide/show metafields (right sidebar to edit if displayed)
- Forestry place metafields **before** the content (and yes sometimes content is not visible by default, you have to scroll down the page to edit body text, especially because they choose to display images if they are defined in YFM)

ashmaroli:
> They might grow too much and fill the whole screen

If that's the only issue, I can have `DocumentEdit.js` and `PageEdit.js` show current style and the `~New` show proposed style. But I don't like that solution personally.

> If we say New Post, do we cover other collections, not just posts, as well?

I didn't consider that possibility. So I'll revert that commit.

mertkahyaoglu:
> The solution might be putting an if statement and change the text according to the collection type.

@ashmaroli What about this? If the currently visited collection type is `posts`, show `New post` else show `New document`.

ashmaroli:
> If the currently visited collection type is posts, show `New post` else show `New document`.

Thats a nice alternative :+1:   
I'll try to implement it..

ashmaroli:
Regarding the metafields placement, I'm still for placing it **before** the content. 
Though I'm open to your experienced input.

/cc @benbalter for an additional opinion.

src/containers/views/Documents.js
```diff
@@ -98,12 +98,19 @@ export class Documents extends Component {
       return null;
     }
 
+    let newBtn = null;
+    if (collection_name == "posts") {
```
mertkahyaoglu: You could use ternary here with `const`.

src/containers/views/Documents.js
```diff
@@ -98,12 +98,19 @@ export class Documents extends Component {
...
+      newBtn = "New Document";
+    }
+
     return (
       <div>
         <div className="content-header">
           <h1>{capitalize(collection_name)}</h1>
           <div className="page-buttons">
-            <Link className="btn btn-active" to={`${ADMIN_PREFIX}/collections/${collection_name}/new`}>New post</Link>
+            <Link className="btn btn-active" to={`${ADMIN_PREFIX}/collections/${collection_name}/new`}>{newBtn}</Link>
```
mertkahyaoglu: If you could add a test [here](https://github.com/jekyll/jekyll-admin/blob/master/src/containers/views/tests/documents.spec.js#L35) for this, it would be great. I'd be glad to help you.

src/containers/views/Documents.js
mertkahyaoglu: And what about naming it `newBtnLabel` ?

ashmaroli: :+1:

src/containers/views/Documents.js
ashmaroli: Feel free to commit to my branch. I need more experience points to modify tests.. :grin:

mertkahyaoglu:
It got out of control when I tried to nest metafields :trollface: 

screen shot 2017-01-13 at 10 37 33 pm

Let's just fix the new button label for this PR and be safe 😄

src/containers/views/Documents.js
mertkahyaoglu: 👍

ashmaroli:
whoa..!
thats interesting..

I'll cherry-pick the button-label-commits onto a new branch..

ashmaroli:
I'll get back to this later. The other commits have been squashed and moved to #274

ashmaroli:
- added a test post with complex Front Matter.
- removed showing list-index-no.s (why was it there anyways ?)
- renamed `new key/value pairs` to `new subkey/value pairs` (the 'parent' metafield key nests subsequent 'subkeys')
- **discovered a new bug:**
  - **When you convert an existing object-value to list-value. the result is invalid YAML**

mertkahyaoglu:
@ashmaroli I appreciate your work but I think we were good with the current design of metafields. This might cause us some unknown design issues in the future as I found some just in a minute like nested dropdowns are not clickable, alignment issues, etc.

screen shot 2017-01-14 at 9 49 18 am

ashmaroli:
okay.. no probs.. any relevant change in the commit list?
I'll close this otherwise.

mertkahyaoglu:
> When you convert an existing object-value to list-value. the result is invalid YAML

What about the bug you've found? It should be reset to empty list.

ashmaroli:
> What about the bug you've found? It should be reset to empty list.

I'm still checking if this is a valid issue.. 
I'll open a dedicated ticket for it.

PR #274, opened 2017-01-13
conditionally alter button-label

ashmaroli:
conditionally alter the label of the button for 'New posts' or 'New documents' based on the current collection.


mertkahyaoglu:
Added the test for new button's label
Lowercase the content type to make it consistent with other buttons (New page, New data file, etc.)

mertkahyaoglu:
Nice team work here @ashmaroli 🎉

ashmaroli:
> Nice team work

@mertkahyaoglu  yeah, I'll try to chip-in more, in the future :smiley: :innocent:

PR #278, opened 2017-01-14
add a tooltip to 'New metadata field' link

ashmaroli:
add a tooltip to let novice users know that metadata will be stored as the Front Matter. :grin:

tooltip

### DIFF
--- a/src/styles/content.scss
+++ b/src/styles/content.scss
@@ -186,4 +186,36 @@
     background: #cfcfcf;
     border: 0;
   }
+
+  .tooltip {
+    position: relative; 
+    
+    .tooltip-text {
+      display: none;
+      position: absolute;
+      top: 36px;
+      right: -250px;
+      width: 300px;
+      padding: 10px;
+      font-size: 1rem;
+      color: #fff;
+      text-align: center;
+      background: #727272;
+      border-radius: 6px;
+
+      &:after {
+        content: "";
+        position: absolute;
+        bottom: 100%;
+        left: 5%;
+        border: 9px solid;
+        border-color: transparent transparent #727272 transparent;
+      }
+    }
+
+    &:hover .tooltip-text {
+      display: block;
+    }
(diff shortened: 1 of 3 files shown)


mertkahyaoglu:
❤️  it!

src/styles/content.scss
```diff
@@ -186,4 +186,36 @@
     background: #cfcfcf;
     border: 0;
   }
+
+  .tooltip {
```
mertkahyaoglu: Could you move this [here](https://github.com/jekyll/jekyll-admin/blob/master/src/styles/metafields.scss#L228)?

src/styles/content.scss
```diff
@@ -186,4 +186,36 @@
...
+      display: none;
+      position: absolute;
+      top: 36px;
+      right: -250px;
+      width: 300px;
+      padding: 10px;
+      font-size: 1rem;
+      color: #fff;
+      text-align: center;
+      background: #727272;
```
mertkahyaoglu: What about `$dark-orange` for background to match it with the link?

src/styles/content.scss
```diff
@@ -186,4 +186,36 @@
...
+      border-radius: 6px;
```
mertkahyaoglu: Let's use `$border-radius` here.

src/styles/content.scss
ashmaroli: was initially there.. moved it to here to denote that its available for all components.

src/styles/content.scss
ashmaroli: I chose `dark-grey` so that its compatible with any other component in future.

mertkahyaoglu: 👍  Then let's store that color as $dark-gray for future uses.

mertkahyaoglu:
Thanks @ashmaroli , really nice addition 👍

PR #280, opened 2017-01-15
Fix encoding bug

### LINKED ISSUE #257: Encoding bug within Editor
I'm running the `server-frontend` script with latest `master` on Windows.
I have a `datafile.yml` within the fixture site's `_data` folder with valid YAML data.

one of the lines being:  
```yaml
name: Jekyll • Simple, blog-aware, static sites
```

As seen when simply opened via `/datafiles`:

ja

As seen on first edit and save:

ja1

As seen after another edit and save:

ja2

mertkahyaoglu:
Fixes #257

@ashmaroli Looks like the problem is that API is reading data files' `raw_content` in a non-utf format. I was thinking that was Ace Editor causing the bug.

/cc @benbalter

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,6 +4,5 @@ source 'https://rubygems.org'
 gemspec
 
 # Site dependencies
-gem 'minima'
 gem 'jekyll-seo-tag'
 gem 'jekyll-sitemap'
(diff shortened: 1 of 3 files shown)


ashmaroli:
cool.. so it wasn't a Windows specific bug either.. Great that you were able to fix it  :+1:

mertkahyaoglu:
>  it wasn't a Windows specific bug either

Actually it was.. I'm not seeing this on my Mac.

I'm guessing `File.read` is not cross-platform?

ashmaroli:
You're right. `File.read` has issues on Windows (especially when its used on binary files). `File.open` with additional parameters, is the preferred alternative.

Issue #186, opened 2016-08-25
Edit drafts


DirtyF:
I'm using the latest version of Jekyll and jekyll-admin on latest OSX / Google Chrome.
### Steps to reproduce
- store a post in `_drafts` folder
- set `show_drafts: true` in`_config.yml`
- Display list of posts in jekyll-admin
- click on a post stored in `_drafts` folder
### Expected

I can edit drafts in jekyll-admin (whether they are stored in a dedcated drafts section or tagged as drafts in the posts section)

Temporary workaround: Don't display drafts amongst posts if they can not be edited  😇 
### Actual

Page : Could not find document 
Notification: Could not fetch the doc

If I edit a post and then try to edit a draft, then the last edited post is displayed instead of the draft, but the notification "Could not fetch the doc" is also displayed as shown in the following screenshot:

![could-not-fetch-document](https://cloud.githubusercontent.com/assets/103008/17952431/c0081d62-6a6a-11e6-9a89-9fc9a2b4e6c5.png)

yoksel:
I have the same problem in the same environment. Editing drafts doesn't working.

mertkahyaoglu:
Related with #150. We need to list directories & files separately for the collection API to be able to show posts in directories. Same approach with PR #244 that @bekicot is working on.

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider whether it can be accomplished in another way. If it cannot, please elaborate on why it is core to this project and why you feel more than 80% of users would find this beneficial.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

iliketoprogram14:
+1

Codeandomx:
+1 Thank you

xow:
Hi. This feature sounds great. Is there any way I can help get this across the line?

mertkahyaoglu:
Hi @xow . We have a PR for this one #367 . You can check it out.

mslinn:
This was supposedly fixed, but jekyll-admin 0.11.0 does not show draft posts with Jekyll 4.2.1.

ashmaroli:
@mslinn Did you enable drafts in your config file by having top-level defn. `show_drafts: true`?
[...]

Issue #150, opened 2016-08-20
Ability to handle collections in subdirectories


benbalter:
Via https://github.com/jekyll/jekyll-admin/issues/136#issuecomment-240958421
### Steps to reproduce:
1. Create `_posts/subfolder/2016-01-01-subfolder-post.md`
2. Visit `/admin/posts`
3. Click the post
### Expected

Ability to edit post
### Actual

Could not find document

To note, `http://localhost:4000/_api/collections/posts/subfolder/2016-01-02-subfolder-post.md` returns the document.

DirtyF:
Same issue if there is a parent folder to sort posts by categories.

Given that I have the following structure: 

```
2016
├── _posts
│   ├── 2015-10-22-une-aventure-humaine.md
│   ├── 2015-12-08-appel-a-orateurs.md
│   ├── 2016-01-08-ouverture-des-inscriptions.md
│   ├── 2016-05-28-curation.md
│   └── 2016-06-20-69-mois-de-sud-web.md
└── index.html
```

Then I should be able to get the posts through the API
Then I should be able to edit posts in jekyll-admin.

dstout:
Also experiencing this issue. This is a great project by the way. Nice work.

virtuald:
Similar issue here, but with collections. I set `permalink: /:slug` in _config.yml for the collection, and then organize the pages underneath in a bunch of directories. I suspect this just means that jekyll-admin is confusing the page slug with the page path, when they aren't necessarily the same thing.

shaunrfox:
+1 I'm using sub-directories for categories on all of the Jekyll sites I've made recently.

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider whether it can be accomplished in another way. If it cannot, please elaborate on why it is core to this project and why you feel more than 80% of users would find this beneficial.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

borisschapira:
Thanks @jekyllbot but I have no data about Jekyll's users that could give me the information I need to evaluate that collection in subdirectories would be beneficial to more than 80% of users.

😢

mertkahyaoglu:
[...]

PR #281, opened 2017-01-19
List files & directories for collection & page APIs

### LINKED ISSUE #163: Can't edit pages in subfolders
I have several pages placed in a `_pages` folder with permalinks manually assigned in the YAML Front Matter.

screen shot 2016-08-21 at 8 54 29 pm

Example: `<site root>/_pages/about.md` with `permalink: /about/`

Admin shows a list of all of them, but when clicking to edit a "Could not find page" error appears.
### Steps to reproduce:
1. Create `test.md` in `_pages`.
2. Add `permalink: /test/` to YAML Front Matter
3. Add `include: ["_pages"]` to _config.yml
4. `bundle exec jekyll serve`
5. Verify `http://localhost:4000/test/` is a valid page.
6. Open Jekyll Admin, navigate to Pages, try to edit `test.md` page

mertkahyaoglu:
Current version of `collection` & `page` APIs list all of the files at the root level. This makes it hard to view/edit files in nested folders in the front end. Other than the user experience, we already have issues like #150, #163, #186 showing that it is not possible to edit files in folders right now. 

This PR proposes listing files & directories all together at a requested level. For example, if we send a request like the following;

**GET** `collections/posts/entries/test`

It returns all directories and files under `test` directory;

screen shot 2017-01-19 at 11 31 48 pm

And the front-end listing view at `posts/test` becomes like this;

screen shot 2017-01-20 at 12 06 22 am

I believe this API model will make it easy to fix aforementioned issues and also provide a better experience for users.

Current status of this PR;

- [x] List files & directories for collections
- [x] List files & directories for pages
- [x] API tests
- [x] Apply changes to the front end
- [x] Front-end tests
- [x] Documentation
- [ ] ~List files & directories for drafts~ (leaving this for another PR)

/cc @benbalter @parkr 

Looking forward to your thoughts on this to move forward confidently.

Closes #163 #150


lib/jekyll-admin/directory.rb
```diff
@@ -0,0 +1,62 @@
...
+  class Directory
+    attr_reader :path
+
+    include Enumerable
+    include JekyllAdmin::URLable
+    include JekyllAdmin::APIable
+
+    TYPE = 'directory'
+
+    def initialize(path, args={})
```
benbalter: I *believe* you're trying to use named arguments here... this can be simplified to:

```ruby
def initialize(path, base: nil, content_type: nil, splat: nil)
  @base = Pathname.new base
  @content_type = content_type
  @splat = Pathname.new splat
  @path = Pathname.new path
end
```

lib/jekyll-admin/directory.rb
```diff
@@ -0,0 +1,62 @@
...
+      {
+        name: name,
+        modified_time: modified_time,
+        path: relative_path,
+        type: TYPE
+      }
+    end
+
+    def name
+      @path.basename
```
benbalter: 1. Since you have `attr_reader :path` above, this can be just `path.basename`

2. To DRY things up a bit, you can add the following in place of the `name` and `modified_time` methods:

```ruby
  extend Forwardable
  def_delegator :@path, :basename, :name
  def_delegator :@path, :mtime, :modified_time
```

lib/jekyll-admin/directory.rb
```diff
@@ -0,0 +1,62 @@
...
+      @path.relative_path_from(@base).to_s
+    end
+
+    def resource_path
+      if @content_type == 'pages'
+        "/pages/entries/#{@splat}/#{name}"
+      else
+        "/collections/#{@content_type}/entries/#{@splat}/#{name}"
+      end
+    end
```
benbalter: Is this method used?

lib/jekyll-admin/directory.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    def url
+      resource_path
+    end
```
benbalter: `alias_method :url, :resource_path`

lib/jekyll-admin/directory.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    def http_url
+    end
```
benbalter: What's your intention here?

lib/jekyll-admin/server/collection.rb
```diff
@@ -59,14 +65,44 @@ def document
...
       def ensure_document
         ensure_collection
         render_404 if document.nil?
       end
+
+      def entries
+        args = {
+          base: site.source,
+          content_type: params['collection_id'],
+          splat: params["splat"].first
```
benbalter: Instead of splat, is this really path?

lib/jekyll-admin/directory.rb
mertkahyaoglu: Since directories do not render to html, I wanted to make it `null` in the resulted JSON.

lib/jekyll-admin/directory.rb
mertkahyaoglu: That's good to know, but I need to pass these arguments [here](https://github.com/jekyll/jekyll-admin/blob/455329fa8d811d0b7a2b11dc032b2a1c7643d75e/lib/jekyll-admin/directory.rb#L58) recursively. Is there a shortcut for that in Ruby?

lib/jekyll-admin/server/collection.rb
mertkahyaoglu: I need this to fill the gap in the `api_url`. For example, if the requested page in `test/other` directory relative to `pages`, I need to be able to generate the `api_url` like this;

/_api/pages/entries/**test/other**/test.md

lib/jekyll-admin/directory.rb
benbalter: `base: base, content_type: content_type, splat: splat`

mertkahyaoglu: These will be undefined since we are outside of `initialize` method.

mertkahyaoglu: I think I should make them `attr_reader`

mertkahyaoglu:
@benbalter I applied the changes you requested, added more tests. 
In addition to them, I changed the `sinatra` routes for pages & collections in order to support directories. Could you review them as well?

For the front end;

- Applied API changes to pages & collections, now you can edit/view the files inside subdirectories
- Added `Title` & `Path` labels to edit view
- Reimplemented `Breadcrumbs` component so that its only task now is to show links. I extracted the input for `path` into another component called `InputPath`.
- Added more tests

The new look;

screen shot 2017-02-02 at 5 12 35 pm


@benbalter I've encountered an issue while renaming files. Sometimes, after writing to the disc, the resulted page/document might be `nil` which causes 404. As far as I see, this happens randomly. Sometimes it renames and returns the renamed file, sometimes it returns nil. So I couldn't figure out what the problem actually is. This is valid for both pages & collections.

@ashmaroli , @DirtyF  Guys, your feedbacks about the user interface are appreciated very much. 😃

lib/jekyll-admin/apiable.rb
```diff
@@ -38,20 +37,26 @@ def to_api(include_content: false, include_documents: false)
...
       if is_a?(Jekyll::Collection)
         output.delete("docs")
-        if include_documents
-          output["documents"] = docs.sort_by(&:date).reverse.map(&:to_api)
-        end
+      end
+
+      if is_a?(Jekyll::Document)
+        output["dir"] = File.join("/", relative_path_from_source.dirname.to_s)
+        output["name"] = relative_path_from_source.basename
```
benbalter: For non-Documents, this should already be calculated as `name`. For documents, this should be already calculated as `basename_without_ext`.

lib/jekyll-admin/apiable.rb
```diff
@@ -38,20 +37,26 @@ def to_api(include_content: false, include_documents: false)
...
+      # array with each rendered document, which we don't want.
       if is_a?(Jekyll::Collection)
         output.delete("docs")
-        if include_documents
-          output["documents"] = docs.sort_by(&:date).reverse.map(&:to_api)
-        end
+      end
+
+      if is_a?(Jekyll::Document)
+        output["dir"] = File.join("/", relative_path_from_source.dirname.to_s)
```
benbalter: I'd add a helper method for this. For non-Documents, this should already be calculated as `@dir`. For Documents, it should be already calculated as `relative_path`.

lib/jekyll-admin/apiable.rb
```diff
@@ -38,20 +37,26 @@ def to_api(include_content: false, include_documents: false)
...
       end
 
       output
     end
 
     private
 
+    # returns the relative path from site's source
+    def relative_path_from_source
+      Pathname.new(file_path).relative_path_from(Pathname.new(site.source))
```
benbalter: I *believe* this should be the value of `@dir` or `relative_path`.

lib/jekyll-admin/directory.rb
```diff
@@ -0,0 +1,58 @@
...
+    def_delegator :@path, :mtime, :modified_time
+    attr_reader :path, :splat, :content_type, :base
+
+    include Enumerable
+    include JekyllAdmin::URLable
+    include JekyllAdmin::APIable
+
+    TYPE = "directory".freeze
+
+    def initialize(path, base: nil, content_type: nil, splat: nil)
```
benbalter: Can you add a comment explaining what each of these arguments represent?

lib/jekyll-admin/directory.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+
+    def to_liquid
+      {
+        :name          => name,
+        :modified_time => modified_time,
+        :path          => relative_path,
+        :type          => TYPE,
+      }
+    end
```
benbalter: Just making a note to be sure to compare this hash to other data types to ensure the field names are consistent.

lib/jekyll-admin/directory.rb
```diff
@@ -0,0 +1,58 @@
...
+    extend Forwardable
+    def_delegator :@path, :basename, :name
+    def_delegator :@path, :mtime, :modified_time
+    attr_reader :path, :splat, :content_type, :base
+
+    include Enumerable
+    include JekyllAdmin::URLable
+    include JekyllAdmin::APIable
+
+    TYPE = "directory".freeze
```
benbalter: It's a strange ruby concept, but strings that aren't modified should be stored as `:symbols` which are treated differently for garbage collection purposes.

lib/jekyll-admin/directory.rb
```diff
@@ -0,0 +1,58 @@
...
+        :type          => TYPE,
+      }
+    end
+
+    def relative_path
+      path.relative_path_from(base).to_s
+    end
+
+    def resource_path
+      if content_type == "pages"
```
benbalter: If we're passing this ourselves, it should be a `:symbol`.

spec/jekyll-admin/server/collection_spec.rb
```diff
@@ -32,56 +32,74 @@ def app
...
-    let(:first_document) { documents.first }
-
     it "returns an individual collection" do
       get "/collections/posts"
       expect(last_response).to be_ok
       expect(last_response_parsed["label"]).to eq("posts")
     end
 
-    it "includes documents" do
+    it "doesn't includes documents" do
```
benbalter: Can you explain the changes to the API that are happening here and why? I'm having trouble following. If we're changing the API do the docs need to be updated?

lib/jekyll-admin/apiable.rb
mertkahyaoglu: Yeap, but `relative_path` includes `filename` while `dir` does not. However, I thought I'd need this in the front end but I didn't. I'm removing this field.

spec/jekyll-admin/server/collection_spec.rb
mertkahyaoglu: I wanted to get your approval before updating them, but sure I can do it right now.

Basically, we are not exposing documents on `/:collection_name` endpoint anymore. This endpoint has only the information about the collection itself as it had in the first release. This could be beneficial in the long term that we might want to change the collection's attributes such as `label` and get only the necessary fields while doing it.

Related to this PR, we now don't have only documents but also have directories that are listed in index endpoints. Therefore putting them all under a field like `documents` or `entries` along with the collection info makes no sense. Instead we put them under their own endpoints. For example;

GET `/posts`
```javascript
{
"label": "posts",
"files": [],
"directory": "/Users/mertkahyaoglu/dev/jekyll-admin/spec/fixtures/site/_posts",
"relative_directory": "_posts",
"permalink": "/:categories/:year/:month/:day/:title:output_ext",
"http_url": null,
"api_url": "http://localhost:4000/_api/collections/posts"
}
```

GET `/posts/entries`, GET `/posts/entries/test`, etc.

```javascript
[
{
"name": "test",
"modified_time": "2017-02-02 17:51:11 +0200",
"path": "_posts/test",
"type": "directory",
...
},
{
"path": "_posts/2016-03-01-test.md",
"id": "/2016/03/01/test",
"relative_path": "_posts/2016-03-01-test.md",
"url": "/2016/03/01/test.html",
"collection": "posts",
...
}
]
```

lib/jekyll-admin/apiable.rb
mertkahyaoglu: Didn't know that. Thanks!

lib/jekyll-admin/apiable.rb
mertkahyaoglu: Just noticed this isn't used anywhere 😄

lib/jekyll-admin/directory.rb
mertkahyaoglu: Good to know that 👍  So I should do this?;

```ruby
TYPE = "directory".to_sym
```

mertkahyaoglu: or directly;

```ruby
{
        :name          => name,
        :modified_time => modified_time,
        :path          => relative_path,
        :type          => "directory",
      }
```

since it is used in only one place.

lib/jekyll-admin/directory.rb
mertkahyaoglu: I based these field names on static files which also have `name`, `modified_time` and `path` fields.

lib/jekyll-admin/directory.rb
mertkahyaoglu: `content_type` passed from `page.rb` as a constant "pages" and from `collection.rb` as a route parameter `collection_id`. Based on the type, `api_url` and `http_url` are generated via `resource_path`. I don't know how this should be `:symbol`

lib/jekyll-admin/directory.rb
benbalter: `TYPE = :directory`

mertkahyaoglu:
> Sometimes, after writing to the disc, the resulted page/document might be nil which causes 404. As far as I see, this happens randomly. 

@benbalter Any idea why this is happening. I'm producing this in older versions too. Could it be related with `site.process` being late to serve new pages/documents?

docs/api.md
```diff
@@ -95,13 +121,15 @@ Delete the requested document from disk.
 * `raw_content` - the page's body (`String`)
 * `front_matter` - the page's YAML front matter (`Object`)
 
-#### `GET /pages`
+#### `GET /pages/:path`
 
-Return an array of page objects. The response does not include the page body.
+Returns an array of pages and directories for the requested path. If `path`
+is not provided, entries at the root level are returned. The response does not include
+pages' body.
```
benbalter: Nice. Does this include directories too (like the collections endpoint)?

lib/jekyll-admin/directory.rb
benbalter: If it's coming from the URL, the this is correct as is. You never want to call `to_sym` on user input, as it's a DDoS vector.

lib/jekyll-admin/server/collection.rb
```diff
@@ -7,32 +7,34 @@ class Server < Sinatra::Base
...
+          delete_file document_path
+          write_path = request_path
         end
 
-        write_file(document_path, document_body)
-        ensure_document
-        json document.to_api(:include_content => true)
+        write_file(write_path, document_body)
+        updated_document = collection.docs.find { |d| d.path == write_path }
+        render_404 if updated_document.nil?
```
benbalter: I *think* you should be able to use `ensure_document` here.

lib/jekyll-admin/server/collection.rb
```diff
@@ -47,26 +49,70 @@ def collection
         collection[1] if collection
       end
 
+      def request_path
+        sanitized_path File.join(site.source, request_payload["path"])
```
benbalter: This should just be `sanitized_path(request_paylod["path"])`. See https://github.com/jekyll/jekyll-admin/blob/140a1f9f598341700796f4044657ea6808f7ea10/lib/jekyll-admin/server.rb#L53-L55.

lib/jekyll-admin/server/collection.rb
```diff
@@ -47,26 +49,70 @@ def collection
...
       end
 
+      def relative_document_path
+        if params["splat"].first.empty?
+          File.join(collection.relative_directory, filename)
+        else
+          relative_to_collection = File.join(params["splat"].first, filename)
+          File.join(collection.relative_directory, relative_to_collection)
+        end
+      end
```
benbalter: How is this being used? We have to be careful here with user input... I could set a filename of `../../../etc/passwd`.

lib/jekyll-admin/server/page.rb
```diff
@@ -32,21 +34,79 @@ class Server < Sinatra::Base
 
       private
 
+      def request_path
+        sanitized_path File.join(site.source, request_payload["path"])
```
benbalter: `site.source` is unnecessary here, as it's already handled in `sanitized_path`

lib/jekyll-admin/server/page.rb
```diff
@@ -32,21 +34,79 @@ class Server < Sinatra::Base
...
+        "#{params["path"]}.#{params["ext"]}"
+      end
+
       def pages
         site.pages.select(&:html?)
       end
 
+      def directory_pages
+        pages.find_all do |p|
+          File.dirname(File.join(site.source, p.path)) == directory_path
```
benbalter: I'd use `sanitized_path` rather than `File.join` here to prevent directory traversal.

lib/jekyll-admin/server/page.rb
```diff
@@ -32,21 +34,79 @@ class Server < Sinatra::Base
...
+        File.join(params["splat"].first, filename)
       end
 
       def page
-        site.pages.find { |p| p.path == params["page_id"] }
+        site.pages.find { |p| sanitized_path(p.path) == page_path }
+      end
+
+      def directory_path
+        sanitized_path File.join(site.source, params["splat"].first)
```
benbalter: Don't need `site.source` here.

benbalter:
> Any idea why this is happening?

Sounds like a race condition. `write_file` calls `site.process`, which shouldn't return until the site is processed... all steps being blocking. Any chance incremental regeneration is on? Should we try disabling it by default?

docs/api.md
mertkahyaoglu: Absolutely 👍  Also fixes #163 since it includes directories like `_pages`.

lib/jekyll-admin/server/collection.rb
mertkahyaoglu: Oh, ok 👍  
The only problem is that this adds extra `/` at the end. Is there a convenient way to delete it?

benbalter: Option 1: Add the slash `path << '/' unless path.end_with?('/')`

Option 2: Remove the slash `path.sub(%r{/$}, '')`

lib/jekyll-admin/server/collection.rb
mertkahyaoglu: I know this doesn't look good but `ensure_document` uses `document` which relies on `document_path` for comparison. We need `write_path` here to compare with.

lib/jekyll-admin/server/collection.rb
mertkahyaoglu: Didn't need that. Translated others to have `/` at the end using `sanitized_path` 👍

mertkahyaoglu: Thanks again 👍  it became very simple now https://github.com/jekyll/jekyll-admin/pull/281/commits/260ddaea8cdeedac1e908578606839150c9615a2

lib/jekyll-admin/server/collection.rb
mertkahyaoglu: Let's say `request_payload["path"]` for `/:collection_id/*?/?:path.:ext` endpoint is `_posts/some_folder/2017-01-01-test.md`. We need to compare this path with the one generated from `params`fields to check if there exists a rename. Any easy solution to achieve this?

mertkahyaoglu:
> Any chance incremental regeneration is on? Should we try disabling it by default?

I tried disabling it from `_config.yml` but no work. Still random errors. Is there any other way to do it?

mertkahyaoglu:
@benbalter Looks like `site.process` is not blocking. I put `sleep(4)` after it and everything worked as expected.

mertkahyaoglu:
@benbalter You're right. I'm opening an issue about that. Thank you for reviewing this PR 👍

Issue #99, opened 2016-08-02
Localization


mertkahyaoglu:
- [ ] Ability to change the language of the front-end

frankis:
Please be aware that i18n is a requirement for many people out there thinking about utilizing jekyll and /admin for their customers. 

Basically I thought /admin would support all configs/settings natively coming with jekyll?

benbalter:
> Basically I thought /admin would support all configs/settings natively coming with jekyll?

Does Jekyll's command line interface localize its output today?

frankis:
Nope, sorry I interpreted "Ability to change the language of the front-end" as a setting to add languages to Posts/Pages from within JekyllAdmin..

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider whether it can be accomplished in another way. If it cannot, please elaborate on why it is core to this project and why you feel more than 80% of users would find this beneficial.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

mertkahyaoglu:
@benbalter is it a good idea to read selected language for the front end from the configuration file? It has to be persistent. Otherwise users will have to change it every time they start Jeykll. Any other option that you might suggest?

oguzzkilic:
I am working on localization infrastructure. I will open a pull-request very soon. :relaxed:

mertkahyaoglu:
@oguzzkilic Awesome 🎉  Thank you for working on this.

erotavlas85:
Any updated on this?

y377:
how to use i18?thank you!

PR #282, opened 2017-01-22
Frontend Improvements

mertkahyaoglu:
This PR mostly aims at code refactoring by creating React components that are reused within the containers.

##  New components
- [Button](https://github.com/jekyll/jekyll-admin/blob/12ffdd1ca8f37b91d7c3afadb6b789420b7af1de/src/components/Button.js): This is a generic component for button elements across the front end. It's responsible for generating button types like `create`, `update`, `view`, `delete` and so on.

- [Errors](https://github.com/jekyll/jekyll-admin/blob/12ffdd1ca8f37b91d7c3afadb6b789420b7af1de/src/components/Errors.js): This one is for displaying validation error messages. It was copy-pasted into a bunch of places before. Making it a component is a good practise.

## Updated Components

- [Breadcrumbs](https://github.com/jekyll/jekyll-admin/blob/12ffdd1ca8f37b91d7c3afadb6b789420b7af1de/src/components/Breadcrumbs.js): Made it a bit more generic to support Data Files since we don't allow changing their filenames, thus their breadcrumbs do not have input field. So adding an `editable` option to the component will allow us to choose whether it should be an input or just a label.

## Localization (One more step to #99)

- To cover all language specific strings, created `lang` folder which contains the files of available languages (only `en.js` for now) and `index.js` file which exports the selected language strings (default to `en` for now).
- `Sidebar` titles and `Button` labels are now imported from `lang` 

##  Other changes

- Removed extra imports such as `brace` import [here](https://github.com/jekyll/jekyll-admin/compare/abstraction?expand=1#diff-4e3a9db16091b30733c3987ad2e4d949L2) which saves us ~300kb in `bundle.js` and some sass imports that have no impact on the front-end style.

- Fixed the issue on Firefox and Safari that the metafield dropdown is not working due to the difference between the behavior of `:focus` css attribute for buttons.

- Finally added Jekyll favicon 🎉

### DIFF
--- a/docs/_frontend/components.md
+++ b/docs/_frontend/components.md
@@ -27,7 +27,7 @@ Can have [all options of SimpleMDE](https://github.com/NextStepWebs/simplemde-ma
 ## Breadcrumbs
 
 Component for generating breadcrumbs. First breadcrumb indicates the current content
-type and the second one which is editable is the path of current item.
+type and the second one which can be either an input or a label is the path of the current item.
 
 ### PropTypes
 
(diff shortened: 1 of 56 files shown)


mertkahyaoglu:
@codegaze You might want to check this [Button](https://github.com/jekyll/jekyll-admin/blob/12ffdd1ca8f37b91d7c3afadb6b789420b7af1de/src/components/Button.js) implementation out. You were working on it back then and we couldn't continue due to conflicts.

codegaze:
@mertkahyaoglu 
Looks great. 
The only thing I'd change is to move these label strings to `constants/messages.js` in favour of #99 🙂

mertkahyaoglu:
@codegaze You're right 👍 . Thanks for the review.

PR #283, opened 2017-01-22
Update appveyor

mertkahyaoglu:


mertkahyaoglu:
@benbalter  I need your help here. Trying to fix #282 appveyor build errors.

mertkahyaoglu:
Closing in favor of #284

benbalter:
Looking into this... for context, `bundle config path vendor/bundle` means that it should be caching the previously downloaded gems to a folder within the checkout, and `bundle install` should use that cache, if the gem's already downloaded, otherwise download the update gem. The install is working fine, but for some reason the tests aren't using that bundle. Looking into it via #284.

Issue #204, opened 2016-08-29
PageCreate can create .txt file but treats it like a .md file


ashmaroli:
### Describe in one or two sentences why you're opening this issue:

I'm new to JekyllAdmin and had no idea that Create New Page should be used to create only markdown documents. Moreover the plugin did not stop me from creating a .txt file.
### Tell us a bit about yourself
- Version of JekyllAdmin I'm using <hint: use `bundle show` to check>: 0.1.0
- Version of Jekyll I'm using <hint: use `bundle show` to check>: 3.2.1
- Operating System <e.g. OS X, Windows>: Windows 7
- Browser <e.g, Safari, Chrome>: Chrome
### Steps to reproduce

Navigate to `http://localhost:3000/admin/page/new`
create demo.txt by entering that in the breadcrumb
Navigate up one level to http://localhost:3000/admin/page
### Expected

Expected demo.txt to not be listed there.
Expected demo.txt to be listed at /staticfiles.
### Actual

demo.txt is listed at /pages
### Other details
- There is no clear indication that tells a user, that this interface is to be used only for creating markdown pages nor any validation if he tries to create a .txt file or .html file. 
- demo.txt created above has the frontmatter dashes inserted.

benbalter:
@ashmaroli to confirm, you created a new page via `page/new`, and you expect it to show up in `staticfiles`, not `pages`? Can you explain a bit more why you expect that? If you created as a new page, shouldn't it be a page? (Pages can have any extension).

ashmaroli:
> Can you explain a bit more why you expect that?

Just for this issue, lets assume, I'm a totally green novice user. ***Hence started this ticket with:***
>I'm new to JekyllAdmin and had no idea that Create New Page should be used to create only markdown documents.

ashmaroli:
Thats how I learned how the plugin differentiates between a page and a static file.
Frontmatter? then Page, else static file.

ashmaroli:
Firstly, I need you to consider this as a bug/a shortcoming within the plugin rather than a rare catch by a novice user.
Secondly, this bug can be fixed by either of the foll.:
 - validating the file extension of the new file to be created. Support only markdown extensions
 - Lock the front end to accept only the filename (w/o extns) and assign a ***.md*** internally (complex solution)

mertkahyaoglu:
@ashmaroli Just tested this and I think this is fixed in v0.2.0. I don't know how though 😬

mertkahyaoglu:
[...]

Issue #205, opened 2016-08-29
Misleading notification when changing file type via JekyllAdmin 


ashmaroli:
### Describe in one or two sentences why you're opening this issue:

Renaming a .txt file as .md file throws a misleading flash notification
### Tell us a bit about yourself
- Version of JekyllAdmin I'm using <hint: use `bundle show` to check>: 0.1.0
- Version of Jekyll I'm using <hint: use `bundle show` to check>: 3.2.1
- Operating System <e.g. OS X, Windows>: Windows 7
- Browser <e.g, Safari, Chrome>: Chrome
### Steps to reproduce

Navigate to `http://localhost:3000/admin/page/new`
create demo.txt by entering that in the breadcrumb
rename demo.txt to demo.md in the breadcrumb
### Expected

Expected demo.txt to be renamed to demo.md silently or with a success! notification
### Actual

demo.txt is successfully renamed to demo.md but with a error flash notification (couldn't update page..)
### Other details
- This could be a false alarm as the plugin is infact rightfully stating that demo.txt couldn't be updated as its no longer available.
- Its nevertheless an alarm that may cause confusion for a short while.
  Possible solution: Introduce a flash notification to handle renaming. This will return a message echoing the old filename with extn and the new filename with extn

mertkahyaoglu:
I can't reproduce this. I think it is a Windows issue. I should definitely test Jekyll Admin in my Windows machine 😕

ashmaroli:
To quote myself ,
> This could be a false alarm as the plugin is infact rightfully stating that demo.txt couldn't be updated **as its no longer available.**

but yes, definitely test on a windows machine as well.

mertkahyaoglu:
As I commented [here](https://github.com/jekyll/jekyll-admin/issues/204#issuecomment-270741477), if you create a page with .txt extension, that page will be created but won't be listed anywhere. So updating it will cause an error because API cannot find it. We should either not allow creating pages with txt extension or not add any front matter if the extension is not markdown. The first one is the simpler I think.

ashmaroli:
> We should either not allow creating pages with `.txt` extension

I think it would be better to not allow change of file extension / type via the Admin interface.

--

> or not add any front matter if the extension is not markdown

don't think that's a good idea. Even `*.html` with Front Matter is a valid JekyllPage object.

mertkahyaoglu:
Closed via #307

PR #301, opened 2017-02-19
Only allow files that render to HTML to be written as a Page

ivantsepp:
Addresses #204 and #205. We are writing files that are considered Jekyll Pages but since #210 change, we should not allow the API to write anything that does not render to HTML

### DIFF
--- a/lib/jekyll-admin/server/page.rb
+++ b/lib/jekyll-admin/server/page.rb
@@ -12,6 +12,11 @@ module JekyllAdmin
       end
 
       put "/*?/?:path.:ext" do
+        unless html_content?
+          status 422
+          content_type :json
+          return json "error_message" => "Invalid file extension for pages"
+        end
         write_path = relative_page_path
         if request_payload["path"] && request_payload["path"] != relative_page_path
           delete_file page_path
@@ -33,6 +38,18 @@ module JekyllAdmin
       end
 
       private
+
+      def html_content?
+        permalink_ext = File.extname(request_payload["front_matter"]["permalink"].to_s)
+        ext = if permalink_ext.empty?
+                File.extname(request_payload["path"] || filename)
+              else
+                permalink_ext
+              end
+        matched_converters = site.converters.select { |c| c.matches(ext) }.sort
+        output_ext = (matched_converters[-2] || matched_converters.last).output_ext(ext)
+        Jekyll::Page::HTML_EXTENSIONS.include?(output_ext)
+      end
 
       def request_path
         sanitized_path request_payload["path"]
--- a/src/utils/fetch.js
+++ b/src/utils/fetch.js
@@ -6,6 +6,7 @@ import {
   getUpdateErrorMessage,
   getDeleteMessage
 } from '../constants/lang';
+import { BadInputError } from './api_errors';
 
 /**
  * Fetch wrapper for GET request that dispatches actions according to the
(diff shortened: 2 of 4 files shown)


ashmaroli:
I haven't tested this branch locally, but if this is case where I try to create a `demo.txt` and the interface tells me `"that's not a valid extension for a `Jekyll::Page"`, cool!
 It'd be cooler, if the message also included the expected extensions that get converted into `.html`

ivantsepp:
Certainly would be nice to get expected extensions that can be rendered into `html` but I'm not sure if that's feasible

src/utils/fetch.js
```diff
@@ -7,6 +7,14 @@ import {
...
 } from '../constants/lang';
 
+function BadInputError(message) {
+  this.name = 'BadInputError';
+  this.message = message || 'Bad input';
+  this.stack = (new Error()).stack;
+}
+BadInputError.prototype = Object.create(Error.prototype);
+BadInputError.prototype.constructor = BadInputError;
+
```
mertkahyaoglu: Could you use ES6 class here and separate it into another file inside `utils` directory ?

ivantsepp: Done!

mertkahyaoglu:
> It'd be cooler, if the message also included the expected extensions that get converted into .html

@ashmaroli Does Jekyll have the list of extensions that are rendered to html? If so, we can append them to the `error_message`.

`Invalid file extension for pages. Only .md, .markdown, .html are supported`

ashmaroli:
> Does Jekyll have the list of extensions that are rendered to html?

By default, Jekyll converts Markdown documents with [these extensions](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/configuration.rb#L26) to `.html` But, the list can expand with the use of [plugins listed here](https://jekyllrb.com/docs/plugins/#converters-1)

mertkahyaoglu:
@ivantsepp Thank you for your time and contribution. This is a great addition. LGTM ❤️ 

@ashmaroli Did you have a chance to test this? I believe this one closes #204 and #205.

ashmaroli:
I tried to test this PR.. but it seems to break the plugin at my end. `test server` is loading any document / page.. dunno if something changed at my end.. If there's no issue at your end, you're welcome to merge this..

ashmaroli:
okay.. it was indeed a bug at my end.. sorted it out..
this PR is clean. :+1:

lib/jekyll-admin/server/page.rb
```diff
@@ -12,6 +12,11 @@ class Server < Sinatra::Base
       end
 
       put "/*?/?:path.:ext" do
+        unless html_content?
+          status 422
+          content_type :json
+          return json "error_message" => "Invalid file extension for pages"
```
benbalter: Looking at the other endpoints, this should be `halt` not `return`

lib/jekyll-admin/server/page.rb
```diff
@@ -12,6 +12,11 @@ class Server < Sinatra::Base
       end
 
       put "/*?/?:path.:ext" do
+        unless html_content?
```
benbalter: I'd move this to a helper method, e.g., `ensure_html_content` to mirror the `ensure_document` and `ensure_directory` helpers.

lib/jekyll-admin/server/page.rb
```diff
@@ -34,6 +39,18 @@ class Server < Sinatra::Base
 
       private
 
+      def html_content?
```
benbalter: Why can't we just [use `html?` here?](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/page.rb#L167-L170) (like we use for `pages` below)? If the file doesn't exist yet, we can simply create the page object, rather than reimplementing potentially divergent logic.

lib/jekyll-admin/server/page.rb
benbalter: I'd make the message more descriptive, e.g., `Files must have an extension of X, Y, or Z`.

lib/jekyll-admin/server/page.rb
ivantsepp: We would need to write the file in order to use `Jekyll::Page` (we could do so via tmp files) but I was avoiding writing any files if it isn't going to be an html file. thoughts?

lib/jekyll-admin/server/page.rb
ivantsepp: Per the discussion above, not sure it's feasible to grab extensions that end up being html files. You can have many converters and they have a method `output_ext(ext)` which takes an ext. We can list all possible extensions

lib/jekyll-admin/server/page.rb
benbalter: Ahh, yes, because the constructor expects the file to exist. Good catch. What about creating a dummy class that includes `convertible`?

ivantsepp:
Closing this in favor of #307 (I deleted my repo so this PR must've lost its reference to my fork)

PR #303, opened 2017-02-20
Improvements to interface

ashmaroli:
* add module name to version `<span>`
* blend logo, sidebar-routes and header
* add icons for file to maintain consistency
* Prepend collection-label with `Collections: ` [debatable]

intefc

### DIFF
--- a/src/components/Editor.js
+++ b/src/components/Editor.js
@@ -34,6 +34,7 @@ class Editor extends Component {
         highlightActiveLine={false}
         className="config-editor"
         fontSize={14}
+        scrollMargin={[15, 15, 15, 15]}
         ref="ace"
         onChange={() => this.handleChange()}
       />
--- a/src/styles/header.scss
+++ b/src/styles/header.scss
@@ -2,20 +2,20 @@
   display: flex;
   align-items: center;
   justify-content: space-between;
-  height: 75px;
-  background-color: white;
+  height: $header-height;
+  background-color: $header-color;
   padding: 0 40px;
-  @include box-shadow(0 2px 4px 0 rgba(204, 204, 204, 0.4));
   .title {
     font-weight: normal;
     a {
-      color: #444444;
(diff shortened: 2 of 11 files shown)


mertkahyaoglu:
I must say there is too much darkness and thickness here that makes me feel bad :trollface:

ashmaroli:
> there is too much darkness and thickness here

thickness?
would reverting the header to white be sufficient?

mertkahyaoglu:
> thickness

I meant bold fonts. Maybe it's just me but really catching my eyes

ashmaroli:
okay.. I'll revert the header to white and reduce the weight on `version`..

src/styles/header.scss
```diff
@@ -2,14 +2,14 @@
   display: flex;
   align-items: center;
   justify-content: space-between;
-  height: 75px;
-  background-color: white;
+  height: $header-height;
+  background-color: $header-color;
+  border-bottom: 1px solid #dedede;
   padding: 0 40px;
-  @include box-shadow(0 2px 4px 0 rgba(204, 204, 204, 0.4));
```
mertkahyaoglu: Let's keep these attributes as they are to be compatible with docs site. It depends but I really love the current design. We can get some feedback from the community to change it.

src/components/Editor.js
```diff
@@ -34,6 +34,7 @@ class Editor extends Component {
         highlightActiveLine={false}
         className="config-editor"
         fontSize={14}
+        scrollMargin={[15, 15, 15, 15]}
```
ashmaroli: this is to simulate padding in the `ace-editor` module. It is activated by an `onScroll` call

ashmaroli: Let me know your decision on this.. try it out locally if you can.. I'll remove the rejected commits and update the branch accordingly..

ashmaroli:
closing in favor of #308

PR #307, opened 2017-02-21
Only allow files that render to HTML to be written as a Page

ivantsepp:
Followup to #301

### DIFF
--- a/lib/jekyll-admin/server/page.rb
+++ b/lib/jekyll-admin/server/page.rb
@@ -12,6 +12,7 @@ module JekyllAdmin
       end
 
       put "/*?/?:path.:ext" do
+        ensure_html_content
         write_path = relative_page_path
         if request_payload["path"] && request_payload["path"] != relative_page_path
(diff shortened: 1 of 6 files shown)


lib/jekyll-admin/server/page.rb
```diff
@@ -34,6 +35,24 @@ class Server < Sinatra::Base
 
       private
 
+      def ensure_html_content
+        unless html_content?
```
benbalter: I believe this can be better implemented as a guard statement, e.g., `return if html_content?`

ivantsepp:
Thanks @benbalter for the pointer to the better implementation!

benbalter:
Nice work @ivantsepp! 🌊  🐙 😛

mertkahyaoglu:
Thanks again @ivantsepp , and @benbalter for helping out ❤️

ivantsepp:
Thanks for looking and reviewing y'all =)

PR #308, opened 2017-02-22
Interface update

ashmaroli:
Cleanup #303 

Interface update:
  - add icons for files to maintain consistency
  - Bump `react-ace` and specify `scrollMargin` array to simulate padding in `Editor` component. (Activated with an `onScroll` function)

Miscellaneous:
  - `import` stylesheets with only "double-quotes"


mertkahyaoglu:
Thanks @ashmaroli 👍

PR #310, opened 2017-02-27
Release v0.4.0

mertkahyaoglu:
Let's release v0.4.0 so that people benefit from new directories api and some other improvements 🎉

### DIFF
--- a/lib/jekyll-admin/version.rb
+++ b/lib/jekyll-admin/version.rb
@@ -1,3 +1,3 @@
 module JekyllAdmin
-  VERSION = "0.3.0".freeze
+  VERSION = "0.4.0".freeze
 end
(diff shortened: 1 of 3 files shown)


ashmaroli:
For some reason, the `scrollMargin` setting from `react-ace-4.1.5` doesn't seem to be working in the release version. But it works perfectly under a `script/server-frontend` session. Please verify.

mertkahyaoglu:
I can't see it in the dev mode either 😕

ashmaroli:
> I can't see it in the dev mode either

is your `react-ace` module at `4.1.5`?

mertkahyaoglu:
Oh, you're right. I'm seeing top margin on dev and nothing on release version, strange 😒

mertkahyaoglu:
I guess I need to reinstall `node_modules` before releasing.

ashmaroli:
> I guess I need to reinstall node_modules before releasing.

:+1:  But I think its only needed when one of the modules get bumped via a PR..

PR #320, opened 2017-03-09
When updating a document or page, read item by itself

parkr:
This prevents a race condition with another thread calling site.process.
Before, site.process had to complete before the response would be
correct, which is quite unreliable. Instead, use the appropriate Jekyll
classes to read the item.

Replaces #294.


lib/jekyll-admin/server/page.rb
```diff
@@ -20,7 +20,9 @@ class Server < Sinatra::Base
         end
 
         write_file(write_path, page_body)
-        updated_page = pages.find { |p| p.path == write_path }
+        updated_page = Jekyll::Page.new(
+          site, site.source, File.dirname(write_path), File.basename(write_path)
```
benbalter: Is the path sanitized prior to being read? (to prevent directory traversal from the user-supplied `write_path`?

lib/jekyll-admin/server/collection.rb
```diff
@@ -29,7 +29,9 @@ class Server < Sinatra::Base
         end
 
         write_file(write_path, document_body)
-        updated_document = collection.docs.find { |d| d.path == write_path }
+        updated_document = Jekyll::Document.new(write_path, {
+          :collection => collection, :site => site,
+        }).tap(&:read)
         render_404 if updated_document.nil?
```
mertkahyaoglu: If we rename `updated_document` to `document`, we can use `ensure_document` which does the same thing. Same for `ensure_page`.

parkr: We won't be able to do this, because `ensure_document` calls `document`, which uses the buggy way of reading:

```ruby
 80       def document
 81         collection.docs.find { |d| d.path == document_path }
 82       end
```

lib/jekyll-admin/server/page.rb
parkr: @benbalter The `write_path` comes from either `document_path` or `request_path`. In my opinion this should most certainly be sanitized before it gets here.

lib/jekyll-admin/server/collection.rb
mertkahyaoglu: Isn't `document` overwritten by `Jekyll::Document.new` ? So `ensure_document` uses newly created document object instead of the result object of `document` method.

mertkahyaoglu: I'm guessing methods' priority is higher over objects in ruby so it will use `document` method anyway?

parkr:
> I think we might need to do this for static files and data files as well.

Static files, yes, but data files use their own API: `DataFile.new(params["data_file_id"])` which reads in the file manually.

lib/jekyll-admin/server/collection.rb
```diff
@@ -78,7 +75,10 @@ def relative_document_path
       end
 
       def document
-        collection.docs.find { |d| d.path == document_path }
+        collection.docs.find { |d| d.path == document_path } ||
+          Jekyll::Document.new(write_path, {
+            :collection => collection, :site => site,
+          }).tap(&:read)
```
mertkahyaoglu: This is really 🆒   You da boss 😎

lib/jekyll-admin/server/page.rb
```diff
@@ -85,7 +84,9 @@ def relative_page_path
       end
 
       def page
-        site.pages.find { |p| sanitized_path(p.path) == page_path }
+        Jekyll::Page.new(
```
mertkahyaoglu: Missed || here or is this on purpose?

parkr: You could do either.

benbalter: I think he's asking if we want to look to the existing page first then fall back to creating a new one like we do for other content types?

lib/jekyll-admin/server/collection.rb
```diff
@@ -29,8 +29,7 @@ class Server < Sinatra::Base
         end
 
         write_file(write_path, document_body)
-        updated_document = collection.docs.find { |d| d.path == write_path }
-        render_404 if updated_document.nil?
+        ensure_document
         json updated_document.to_api(:include_content => true)
```
mertkahyaoglu: This needs to be `document` now instead of `updated_document`.

lib/jekyll-admin/server/page.rb
```diff
@@ -20,8 +20,7 @@ class Server < Sinatra::Base
         end
 
         write_file(write_path, page_body)
-        updated_page = pages.find { |p| p.path == write_path }
-        render_404 if updated_page.nil?
+        ensure_page
         json updated_page.to_api(:include_content => true)
```
mertkahyaoglu: This needs to be `page` too.

mertkahyaoglu:
This should be good to go after deleting test files.

parkr:
> This should be good to go after deleting test files (unless failing tests generating them).

@mertkahyaoglu Would you mind taking this PR and running with it? I am not able to commit any more time to it unfortunately but I think we have the right approach here. 😄

mertkahyaoglu:
@parkr Got it 👍 

I reverted 1203e14 because `write_path` becomes undefined in `page`, `document` methods.

Another thing is that static files tests are failing when `JekyllAdmin::StaticFile.new` is used.

Here are the failing tests;

```shell
rspec ./spec/jekyll-admin/server/static_file_spec.rb:30 # static_files 404s when a static file doesn't exist
rspec ./spec/jekyll-admin/server/static_file_spec.rb:35 # static_files returns a deep directory listing
rspec ./spec/jekyll-admin/server/static_file_spec.rb:51 # static_files returns a directory listing starting only at the root
```

@benbalter Any solution for this? Or shall we revert this approach for static files.

benbalter:
@mertkahyaoglu: @parkr is out of the office this week, but I'll take a look this morning and see what I can do here to get the build passing.

lib/jekyll-admin/server/page.rb
```diff
@@ -75,25 +61,13 @@ def directory_paths
...
-        sanitized_path params["splat"].first
-      end
-
-      def ensure_directory
-        render_404 unless Dir.exist?(directory_path)
+        found = site.pages.find { |p| sanitized_path(p.path) == path }
+        return found if found
+        return unless File.file?(write_path)
+        Jekyll::Page.new(
+          site, site.source, File.dirname(write_path), File.basename(write_path)
```
mertkahyaoglu: Do we want the full path for `dir` ?

benbalter:
I started digging in to how we were reading and writing files, and I realized we implemented it four different ways for each content type. This led to a few problems:

* There was a lot of duplicate, but slightly different code (with slightly different behaviors)
* Sometimes `path` was relative, sometimes it was relative with a preceding `/`, sometimes it was absolute
* There were several paths where user input was not sanitized before reading from and writing to disk 
* Some data formats could not be renamed (data files, static files)
* The API would not 404 if you tried to rename a file that didn't exist

As a result now:

* Created new `PathHelper` and `FileHelper` mixins to break path and file logic out of `Server`
  * `FileHelper` now exposes `requested_file` and `written_file` methods, as well as `ensure_requested_file` and `ensure_written_file`. The "requested file" is the file in the URL. The "written file" is what we actually wrote to disk. It may be the same as the requested file, or may be different if the file is renamed. This takes the place of the `page`, `document`, etc. helper methods.
  * `PathHelper` exposes consistent `absolute_path`, `relative_path`, `absolute_write_path`, and `relative_write_path` helpers along with `renamed?` to determine if a given request is renaming a file
  * `DataFile` now relies on `PathHelper` to ensure consistent sanitization and handling of paths (relative URLs now have preceding `/`s)
* All data types can now be renamed and will properly 404 if you try to rename an invalid file
* The collection hash now exposes an `entries_url` field (useful for debugging)
* Lazy load child classes (minor performance improvement that cleans up the main file)

As for the original race condition, I believe this PR should fix it with one gotcha:

I could not figure out a way to get notified when the parent watch process finishes (and thus we should read the file and respond to the request). Reading the file directly resulted in the raw content, but we still would have to run the read and generate steps, which given plugin, I couldn't properly reimplement per-file.

Instead, JekyllAdmin disables the `--watch` and `--detach` flags, and will regenerate the site in line in its own process, preventing the race condition whereby the file would be read between `Site#reset` and `Site#read`.

The only gotcha is that the server will not automatically regenerate for files edited outside of JekyllAdmin (e.g., CSS or Javascript). One solution here, might be to create a process endpoint, which a frontend could create a button for.

@mertkahyaoglu there's a lot here (I fell down the 🐰 hole), but let me know if it fixes the original race condition?

mertkahyaoglu:
@benbalter This is a great refactoring 💯  Clean and readable code even for a Ruby newby like myself 😄 Thanks for the great effort 👍 

One thing is that the race condition is still present on my side since `watch` is not disabled.

mertkahyaoglu:
> One solution here, might be to create a process endpoint, which a frontend could create a button for.

This makes sense to me. Could place a button on `configuration` page which says `Regenerate Site`.

benbalter:
> One thing is that the race condition is still present on my side since watch is not disabled.

@mertkahyaoglu helps to actually include the file. Try 5a87555.

mertkahyaoglu:
> Try 5a87555

@benbalter Sad to show you this 😭 

screen shot 2017-03-14 at 10 39 22 pm

`watch` is still alive.

ashmaroli:
The `options["watch"` actually comes to play in [`commands/build.rb`](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/commands/build.rb#L40-L42). For some reason, I was able to **disable `watch` only by patching `Build#build`**. Surprisingly, patching the `Build#process` **had no effect**

lib/jekyll/command.rb
```diff
@@ -0,0 +1,22 @@
+module Jekyll
+  class Command
+    class << self
+      JEKYLL_ADMIN_OVERRIDES = {
+        "watch"    => false,
+        "detached" => false,
```
ashmaroli: this should simply be `detach`, no?

benbalter: Yes, but in hindsight, it's not needed as it would detach both server processes, so we should be fine.

benbalter:
@ashmaroli Wow. Great catch. Implemented via e26934f.

I also broke the refactor out into its own PR via https://github.com/jekyll/jekyll-admin/pull/323 to keep things distinct. If that looks good, we can merge #320, and then if tests pass here, and @mertkahyaoglu confirms this resolves the race condition, I believe the only thing remaining here would be to document the limitation.

ashmaroli:
Thanks @benbalter :smiley: .. but the tests don't seem too happy about this change.

mertkahyaoglu:
Shall we change the title of this PR?

benbalter:
@mertkahyaoglu does this PR fix the problem for you locally? If so, I may just remove the test... I've spent way too much time trying to figure out how to test this.

mertkahyaoglu:
Thanks @benbalter, this PR fixes the problem 🎉

mertkahyaoglu:
For the test, can we write to a file manually then expect the regeneration to not happen? I don't know how we can check whether the regeneration is happening doing that. Maybe we use a `hook` and expect it to be fired only once instead of twice as it was before this PR.

I have no clear solution, just throwing some ideas.

spec/jekyll-admin/integration_spec.rb
```diff
@@ -44,4 +45,10 @@
       expect(response.body).to match("collections_api")
     end
   end
+
+  context "watching" do
+    it "Jekyll isn't watching" do
+      expect(@stdout.read).to include("Auto-regeneration: disabled")
```
mertkahyaoglu: I see `Incremental build: disabled` in the console. Are these different?

ashmaroli: yes. they are..
incremental build can be enabled with a `--incremental` switch

ashmaroli:
There's a bug with the current implementation. The site does regenerate with `--watch` even though the terminal / stdout says its disabled

mertkahyaoglu:
> The site does regenerate with --watch even though the terminal / stdout says its disabled

How is that happening? Could you explain please?

I'm not seeing any regeneration logs in the console after saving a file from my editor. I also renamed files many times from the panel and there were no errors like before.

ashmaroli:
> How is that happening?

Perhaps its the `--watch` switch to `npm` because this occurs only while testing with `server-frontend` and does not occur with just `test-server`

ashmaroli:
can i send a PR to this branch?

benbalter:
> can i send a PR to this branch?

Please.

ashmaroli:
> For the test, can we write to a file manually then expect the regeneration to not happen? I don't know how we can check whether the regeneration occurs doing that.

sent a PR #324  inspired by this comment.

benbalter:
Full props go to @ashmaroli. If tests are green, I believe this can be merged.

mertkahyaoglu:
@ashmaroli @benbalter @parkr Awesome work guys. Many thanks. This and #323 deserve 0.4.1 release.

ashmaroli:
Thank you for the acknowledgement @benbalter and @mertkahyaoglu :smiley: :beers:

ashmaroli:
> This and #323 deserve a release.

I feel #321 should be addressed by the next release. Either with the merge of #322 (or another better implementation), **or** an entry to the documentation informing users about the situation.

Another TODO: 
> I believe the only thing remaining here would be to document the limitation.

from Ben's comment [above](https://github.com/jekyll/jekyll-admin/pull/320#issuecomment-286751125).

mariusa:
This introduces other issues, see #340 

Not sure why the admin has to take over re-generation when enabled as a plugin? Just write files to disk, like all editors. Then jekyll would do it's job and re-generate content on changes, no matter if files are changed via admin or other editors. 

Thanks

ashmaroli:
From Jekyll Admin's [online docs](https://jekyll.github.io/jekyll-admin/architecture/)

> **Note:** Since there are two Sinatra servers that might call `site.process` concurrently, Jekyll Admin disables `--watch` flag to prevent a race condition between these servers that might cause incorrect responses for the API.
This ensures that the site is regenerated by only the process that Jekyll Admin runs.

mariusa:
The current editing of .css, layouts or includes is a pain. We have to stop jekyll and start again on each css edit, to see changes.
What solutions do you recommend?

The simplest one does see to limit jekyll-admin to edit files as any IDE, just write them to local disk. Don't try to generate anything, jekyll already handles that well, and it works for all assets.

ashmaroli:
The simplest solution when editing "non-content files" is to comment out `jekyll-admin` from your `_config.yml` or `Gemfile`  depending on how its being used..

mariusa:
So
1. edit some pages in jekyll-admin
2. realize we need to edit some CSS/JS/include
3. stop jekyll
4. comment out jekyll-admin
5. start jekyll
6. multiple CSS edits
7. stop jekyll
8. uncomment jekyll-admin
9. start jekyll
10. back to page editing

Would you agree it's cumbersome?

ashmaroli:
> Would you agree it's cumbersome?

Definitely.
I'm planning to release a gem-version of my fork of `jekyll-admin` soon.. It'll support editing css and other non-content files via the interface.. (still disables `--watch` though..)

mertkahyaoglu:
Actually I'm planning to bring the `watch` back. With the changes I described [here](https://github.com/jekyll/jekyll-admin/issues/289#issuecomment-310740305), the possibility of  hitting a race condition is very low (%0 for me, please test #415 locally and let me know if you encounter a race condition. It will work in `dev` mode for you to test). If the possibility is acceptable, I think we should give it a shot.

ashmaroli:
> I'm planning to release a gem-version of my fork

FYI, released under gem-name `jekyll-manager`

PR #322, opened 2017-03-12
Support custom config files from CLI

### LINKED ISSUE #321: auto selecting jekyll config file
### Description:
Jekyll-admin does not load the configuration file selected for start Jekyll.

* Version of JekyllAdmin I'm using : 0.4.0
* Version of Jekyll I'm using : 3.4.1
* Operating System : Windows 10 x64
* Browser : Waterfox 

### Steps to reproduce:
I'm using two Jekyll config files: a _config_dev.yml for local testing and _config.yml for GitHub pages, in that way; i can tinkering with jekyll without mess the live web page.

I use this command for the local testing:

```cmd
jekyll serve --config _config_dev.yml --drafts --incremental
```
then i put in my browser

```html
http://localhost:4000/admin/configuration
```

### I expected the following:
jekyll-admin show the content of the config file selected in jekyll: _config_dev.yml

jekyll-admin use as configuration reference _config.yml instead the one i mark for run jekyll :confused:

ashmaroli:
Fixes #321 

When you provide an array of config files via the `--config` switch, we'll read and write to the first of such files.

- [x] add code
- [ ]  test code

### DIFF
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -2,6 +2,21 @@ module Jekyll
   module Commands
     class Serve < Command
       class << self
+        def process(opts)
+          @overrides = opts["config"] if opts["config"]
+
+          opts = configuration_from_options(opts)
+          destination = opts["destination"]
+          setup(destination)
+
+          start_up_webrick(opts, destination)
+        end
+
+        def custom_configs
+          return nil unless @overrides
+          @overrides
+        end
+
         def start_up_webrick(opts, destination)
           server = WEBrick::HTTPServer.new(webrick_opts(opts)).tap { |o| o.unmount("") }
           server.mount(opts["baseurl"], Servlet, destination, file_handler_opts)
--- a/lib/jekyll-admin/server/configuration.rb
+++ b/lib/jekyll-admin/server/configuration.rb
@@ -28,9 +28,17 @@ module JekyllAdmin
         configuration.read_config_file(configuration_path)
       end
 
+      def custom_configs
+        Jekyll::Commands::Serve.custom_configs
+      end
+
       # Returns the path to the *first* config file discovered
       def configuration_path
-        sanitized_path configuration.config_files(overrides).first
+        if custom_configs
+          sanitized_path custom_configs.first
+        else
+          sanitized_path configuration.config_files(overrides).first
+        end
       end
 
       # The user's uploaded configuration for updates


ashmaroli:
um.. I can't figure out how to emulate CLI with current test-suite..

lib/jekyll/commands/serve.rb
```diff
@@ -2,6 +2,21 @@ module Jekyll
   module Commands
     class Serve < Command
       class << self
+        def process(opts)
```
benbalter: Rather than monkey patching another method, can we pass the options hash to our existing monkeypatch?

lib/jekyll/commands/serve.rb
```diff
@@ -2,6 +2,21 @@ module Jekyll
...
+          opts = configuration_from_options(opts)
+          destination = opts["destination"]
+          setup(destination)
+
+          start_up_webrick(opts, destination)
+        end
+
+        def custom_configs
+          return nil unless @overrides
+          @overrides
```
benbalter: Could this be simplified to an `attr_accessor`?

lib/jekyll/commands/serve.rb
ashmaroli: I couldn't find another way to bypass this [deletion](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/configuration.rb#L157)

lib/jekyll/commands/serve.rb
ashmaroli: I was under the impression that when we define an `attr_accessor`, its called again via `#initialize` method

ashmaroli:
> What's the desired user experience here?

**Primarily, to be in sync with existing Jekyll functionality and be able to *edit* the site built/served with the specified configuration file.**

--

> To only be able to modify the first configuration file? Would it make more sense to abstract configuration into configuration files?

**When the user builds/serves via a list of config files, its [understood and documented](https://jekyllrb.com/docs/configuration/#build-command-options) that the effective config hash will be a merged singularity. But instead of displaying contents of all the config files in the list, we shall simply allow writing to the first of specified config files.**

--

> How often do we think users will want to modify secondary configuration files?

**Jekyll's build options allow one to specify custom config files. JekyllAdmin's UX gets hampered when we turn a blind-eye to an existing long-standing core functionality.**

>  IIRC having a singular configuration file was a purposeful design decision to keep things slim (but that doesn't mean it can't be changed now if there's a strong reason).

**For initial release, that was a wise decision. But its time to expand the scope..** :grin:

ashmaroli:
I believe this is ready a final review..

ashmaroli:
*Update: This will not work on `Jekyll-3.5.0`. Expecting refinements from Jekyll itself..*

ashmaroli:
@mertkahyaoglu @benbalter Guys, I believe this is once again ready for a scrutiny..
I temporarily disabled a Rubocop check to have it handled as required in a separate PR..

PR #324, opened 2017-03-15
test --watch with this plugin

ashmaroli:
fix integration spec for testing `--watch` with `JekyllAdmin`

subset of #320

### DIFF
--- a/spec/jekyll-admin/integration_spec.rb
+++ b/spec/jekyll-admin/integration_spec.rb
@@ -10,8 +10,7 @@ describe "integration" do
   let(:response) { Net::HTTP.get_response(uri) }
 
   before do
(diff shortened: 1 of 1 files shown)


wraybowling:
Not being able to --watch files really threw me off for a number of weeks. I wonder if there's a way we can have both. :/

agjs:
Is this going to be fixed soon ?

mertkahyaoglu:
@agjs Hopefull 😄  If we can find a decent solution..

agjs:
@mertkahyaoglu Gotcha. Will soon start contributing since we'll need some custom editor features.

mertkahyaoglu:
@agjs Awesome, glad to have you here ❤️

PR #325, opened 2017-03-16
Translated current English language constants to Danish

BlueWombat:
I haven't actually been able to test it yet, and I know it might be a bit premature, since it seems no language selection feature exists as of yet.

### DIFF
--- a/src/constants/lang/da.js
+++ b/src/constants/lang/da.js
@@ -1,0 +1,77 @@
+// inline messages
+export const getDeleteMessage = (filename) =>
+  `Er du sikker på du ønsker at slette "${filename}" ?`;
+
+export const getLeaveMessage = () =>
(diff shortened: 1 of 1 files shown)


mertkahyaoglu:
It is actually good to translate earlier so we have a reason to implement language selection feature. Thanks for starting this 👍  We might need your support in the future for additional translations.

BlueWombat:
Great, you're welcome :-)

Well, Danish is my native tongue, so that was a n easy one. I might also be able to help out with Swedish and Norwegian, because though I'm not fluent, I should still be able to make a meaningful translation.

I also plan on contributing with actual code, just don't have a ton of time right now, as I'm in the middle of moving :-/

mertkahyaoglu:
> I might also be able to help out with Swedish and Norwegian

That's perfect 👍 

>I also plan on contributing with actual code, just don't have a ton of time right now

Contribute whenever you want. It's greatly appreciated ❤️

PR #326, opened 2017-03-16
add locale FR-fr

DirtyF:
Try to switch locale to french, but tests are not passing as some are verifying the messages. 😞


mertkahyaoglu:
It's ok. We will fix those tests once we implement the switcher.

Thanks for the translation @DirtyF 👍

PR #331, opened 2017-03-22
Document #320

mertkahyaoglu:


ashmaroli:
I'm kinda confused now, where did we disable `--detach` flag?  
And how does the site regenerate when `--watch` has been disabled?

mertkahyaoglu:
> where did we disable --detach flag?

You're right. It was from @benbalter's first implementation. Thought we disabled it.

> how does the site regenerate when --watch has been disabled?

We call `site.process` after writing to a file.

ashmaroli:
> It was from @benbalter's first implementation. Thought we disabled it.

Is it necessary that `--detach` be disabled? Note: You were not able to reproduce the race condition with current implementation.


> We call `site.process` after writing to a file.

Ah yes! That true. Should add that to our tests if its not already tested.

PR #337, opened 2017-03-31
Support frontmatter defaults in the front end

### LINKED ISSUE #165: Allow for default fields
Following from #140, I'm breaking this issue out. 

Allow for default fields. A site owner should be allowed to set up fields in the config file, or elsewhere, which creates consistency and ease of use for content creators:

Here are examples for how others handle this: [Prose.io:](https://github.com/prose/prose/wiki/Prose-Configuration#metadata-configuration),  [Cloudcannon](https://docs.cloudcannon.com/editing/collection-defaults/), [Siteleaf](http://learn.siteleaf.com/content/defaults/).

One approach, as mentioned by @benbalter, is that I could have the following in my config:

```
jekyll_admin:
  default_fields:
    - title
    - description

```

And then when creating a new post, those fields would be initialized, but empty, by default. This is closest to the [prose.io configuration](https://github.com/prose/prose/wiki/Prose-Configuration#metadata-configuration).

mertkahyaoglu:
Ability to view frontmatter defaults in the front end. Frontmatter defaults are inserted into metadata section of pages/documents' views to let users visually see all of the fields they have.

To achieve this, we read frontmatter defaults from the config file and inject them into metadata of pages and documents.

Fixes #165


src/containers/views/DocumentEdit.js
```diff
@@ -113,6 +114,8 @@ export class DocumentEdit extends Component {
     } = currentDocument;
     const [directory, ...rest] = params.splat;
 
+    const metafields = injectDefaultFields(config, directory, collection, front_matter);
```
ashmaroli: `config` is since recently, 
```
{"content": {"foo": "bar"}, "raw_content": "foo: bar\n"}
```
So **either** this line and all instances of `config` passed to `injectDefaultFields` should be changed to `config.content` **or** change `injectDefaultFields` function to extract `defaults` from `config.content`

src/components/metadata/MetaSimple.js
```diff
@@ -76,7 +76,7 @@ export class MetaSimple extends Component {
         <TextareaAutosize
           onChange={(e) => this.handleEditableChange(e)}
           className="field value-field"
-          defaultValue={fieldValue}
+          value={fieldValue}
```
ashmaroli: this needs interpolation to convert `Boolean` to literal.. (detected via Browser Console)
```
value = {`${fieldValue}`}
```

ashmaroli:
@mertkahyaoglu found a minor bug with this via `server-frontend`
If you were to access a `page` for `editing`, the browser console shows an  momentary error due to a delay in fetching the `config.content` I'll open a dedicated ticket if you want..

mertkahyaoglu:
Thanks @ashmaroli 👍 . Recently I came across with [this library](https://github.com/facebookincubator/idx) from Facebook that will fix this kind of console errors very easily. I will try to integrate this when I get some time.

PR #339, opened 2017-04-07
Strip trailing and leading whitespace in Metafield keys and values

### LINKED ISSUE #338: Strip spaces after metadata field label
### Description:
If a user adds a space at the end of a metadata field label, it turns the front matter into a string and causes it not to work. I release they can just remove the space in the admin, but my users are humans that make mistakes sometimes and don't know the intricacies of jekyll admin like I do. #

[Screenshot of metadata field with space at end](http://i.imgur.com/o9wxWR0.png) leads to this in the front matter: `'status ': live`

Jekyll Admin v0.4.1

### Steps to reproduce:

1. Create a new document
2. Add a new metadata field
3. Enter a label for the new field, then add a space at the end
4. Enter content for the new field
5. Select Save
6. Look at the new md document in your repo

You can also try to load your content with liquid: ex {{ page.new_field }}

### I expected the following:
I would like for Jekyll Admin to remove extra spaces at the end/beginning if they exist. 

I get a unusable front matter item that messes up my whole operation

ashmaroli:
fix #338 
opted to strip metafield values as well

### DIFF
--- a/src/actions/metadata.js
+++ b/src/actions/metadata.js
@@ -33,6 +33,7 @@ export function updateFieldKey(namePrefix, fieldKey, newKey) {
 }
 
 export function updateFieldValue(nameAttr, value) {
+  value = value.trim();
   return {
     type: ActionTypes.UPDATE_FIELD_VALUE,
     nameAttr,
--- a/src/actions/metadata.js
+++ b/src/actions/metadata.js
@@ -23,6 +23,7 @@ export function removeField(namePrefix, key) {
 }
 
 export function updateFieldKey(namePrefix, fieldKey, newKey) {
+  newKey = newKey.trim();
   return {
     type: ActionTypes.UPDATE_FIELD_KEY,
     namePrefix,
--- a/src/actions/metadata.js
+++ b/src/actions/metadata.js
@@ -23,7 +23,6 @@ export function removeField(namePrefix, key) {
 }
 
 export function updateFieldKey(namePrefix, fieldKey, newKey) {
-  newKey = newKey.trim();
   return {
     type: ActionTypes.UPDATE_FIELD_KEY,
     namePrefix,
@@ -33,7 +32,6 @@ export function updateFieldKey(namePrefix, fieldKey, newKey) {
 }
 
 export function updateFieldValue(nameAttr, value) {
-  value = value.trim();
   return {
     type: ActionTypes.UPDATE_FIELD_VALUE,
     nameAttr,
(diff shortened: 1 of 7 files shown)


src/actions/metadata.js
```diff
@@ -23,6 +23,7 @@ export function removeField(namePrefix, key) {
 }
 
 export function updateFieldKey(namePrefix, fieldKey, newKey) {
+  newKey = newKey.trim();
```
mertkahyaoglu: Let's do the trimmings in the reducers.

mertkahyaoglu:
Rubocop rules are changing so often :trollface:

ashmaroli:
> Rubocop rules are changing so often

The easy fix would be to lock to a certain minor version and bump only if it provides significant performance gains
```ruby
gem "rubocop", "~> 0.48.1"
```

ashmaroli:
@mertkahyaoglu Should I create a PR to fix and lock Rubocop or will you push a commit to `master` and restart the tests here?

mertkahyaoglu:
We have `spec.add_development_dependency "rubocop", "~> 0.35"` in `jekyll-admin.gemspec`. What's that for?

ashmaroli:
> We have spec.add_development_dependency "rubocop", "~> 0.35" in jekyll-admin.gemspec. What's that for?

* Rubocop is a dependency only during developing the plugin
* `~>0.35` means use a version `>= 0.35 < 1.0`

mertkahyaoglu:
> ~>0.35 means use a version >= 0.35 < 1.0

Oh, got it now. It means equal to or greater than in the last digit. Very different than the one for JS community 😬

ashmaroli:
> Very different than the one for JS community

How is it in the JS community?

mertkahyaoglu:
Quoted from stack overflow;

```
In the simplest terms, the tilde matches the most recent minor version (the middle number). ~1.2.3 will match all 1.2.x versions but will miss 1.3.0.

The caret, on the other hand, is more relaxed. It will update you to the most recent major version (the first number). ^1.2.3 will match any 1.x.x release including 1.3.0, but will hold off on 2.0.0.
```

ashmaroli:
Ah! yes, I somewhat realized that from this plugin's `package.json` :grin:

ashmaroli:
for rubygems,
`~> 1.15` means `>= 1.15 < 2.0` (lock major)
`~> 1.35.3` means `>= 1.35.3 < 1.36.0` (lock minor)

ashmaroli:
@mertkahyaoglu why isn't the trimming working for the spec fixture, while manually testing it via the frontend seems to work fine as intended?

PR #342, opened 2017-04-10
ES6 applied in some files

gkal19:


mertkahyaoglu:
Thanks @gkal19 👍  
Great to have you here as a first-time contributor 😄

Issue #209, opened 2016-08-31
Link theme files via JekyllAdmin interface


ashmaroli:
I request to have theme files linked via JekyllAdmin.

Say, I have a blog setup with Jekyll 3.2.1 on Windows
In order to change the stylesheets or layouts or even a minor change to one of the includes, I've to open another Explorer window and navigate to my gems folder and then jump back-and-forth between the various directories as I change things around. Novice user or advanced, this is quite irritating.

Now that Jekyll sports a UI, it should automatically show theme files ( and be able to edit them ). Syntax-highlighting is a plus, not necessary, at the moment.

Thanks :smiley:

benbalter:
@ashmaroli thanks for the suggestion. For context, when @mertkahyaoglu and I were originally specing out the project, we purposely placed theme/layout editing out of scope for three reasons:

1. The problem we were trying to solve was authoring and managing *content*, and we wanted to focus on the content editing interface, so that we could make it awesome.

2. Editing a layout is very different than editing a markdown file. While it's easy to add a Markdown editor to a view, there are many purpose-built tools out there that do a much better job at editing HTML (e.g. Atom).

3. The type of user that could most benefit from Jekyll Admin, are those that don't want to edit themes or layouts. They just want to edit content. Those interested in editing layouts are generally more advanced users, and can be expected to install a dedicated text editor, which will be better suited for the task.

ashmaroli:
@benbalter, okay. the initial scope excluded themes on purpose. I get that.
Can the scope be extended, and have this FR under consideration for the near future? :smiley:

DirtyF:
I'd rather see the project focus on enhancing the writing user experience (think keyboard shortcuts, better typography for instance, customizing editor to your needs) than trying to mimic Wordpress and clutter the UI. [Removing a feature is a feature](http://ben.balter.com/2016/07/21/removing-a-feature-is-a-feature/).

mertkahyaoglu:
Agree with @DirtyF , @benbalter. Sorry took so long 😬

PR #344, opened 2017-04-11
API support for gem-based themes

ashmaroli:
Related to #209
Re-proposing support for gem-based themes.

If this is still beyond the scope of this plugin, can it be released as an extension of `JekyllAdmin` (d'oh plugin for a plugin...)

--
/cc @benbalter @mertkahyaoglu


lib/jekyll-admin/server/theme.rb
```diff
@@ -0,0 +1,117 @@
...
+      def ensure_theme
+        render_404 unless site.theme
+      end
+
+      def at_root(dir = "")
+        site.in_theme_dir(dir)
+      end
+
+      def theme
+        site.theme.send(:gemspec)
```
benbalter: I'm a bit hesitant to use a private API here. Why can't we just use the public API (which exposes name, version, etc.)?

lib/jekyll-admin/server/theme.rb
```diff
@@ -0,0 +1,117 @@
+module JekyllAdmin
+  class Server < Sinatra::Base
+    namespace "/theme" do
+      get do
+        ensure_theme
+        json(
+          {
+            :name        => theme.name,
+            :version     => theme.version,
+            :authors     => theme.authors.join(", "),
```
benbalter: What's the value of exposing the authors and license here?

lib/jekyll-admin/server/theme.rb
```diff
@@ -0,0 +1,117 @@
...
+            :license     => theme.license,
+            :path        => at_root,
```
benbalter: Will this be the path to the gem root (outside the source directory)?

lib/jekyll-admin/server/theme.rb
```diff
@@ -0,0 +1,117 @@
...
+            :directories => entries.map(&:to_api),
+          }
+        )
+      end
+
+      get "/*?/?:path.:ext" do
+        ensure_theme
+        json(
+          { :name => filename,
+            :path => file_path, }.merge!(contents).merge!(url_fields)
```
benbalter: What's going on with the merges here?

lib/jekyll-admin/server/theme.rb
```diff
@@ -0,0 +1,117 @@
...
+      end
+
+      private
+
+      def ensure_theme
+        render_404 unless site.theme
+      end
+
+      def at_root(dir = "")
+        site.in_theme_dir(dir)
```
benbalter: Why do we need a helper here? If we do, can we call it `in_theme_dir` or something similar for consistency (rather than introduce a new naming convention)?

lib/jekyll-admin/server/theme.rb
```diff
@@ -0,0 +1,117 @@
...
+          }
+        end
+      end
+
+      def splats
+        params["splat"] || ["/"]
+      end
+
+      def layouts(file)
+        Jekyll::Layout.new(site, directory_path, file)
```
benbalter: This is `layouts` plural, but appears to return a single `Layout`?

ashmaroli: You've got a valid point here. Will change the method.

lib/jekyll-admin/server/theme.rb
ashmaroli: The Public API *only exposes* `name` and `version`. This method surfaces all other possible data (`licence`, `author`, `dependencies`, etc)

lib/jekyll-admin/server/theme.rb
ashmaroli: merge base hash with additional hashes from respective methods and pass the resulting hash to be output as API points

lib/jekyll-admin/server/theme.rb
ashmaroli: okay. Will use `site.in_theme_dir` directly instead of a wrapper

ashmaroli:
> Would love to hear more about your use case here. Not opposed to exposing theme metadata, but not sure that we should support theme authoring via Jekyll Admin as a core use case

Intention:
  * Expose data from the theme-gem in use with **`GET`** and display it via the front end.
    * Dedicated `api_url` for each template file. Define `http_url` only for rendered files (assets in `_site`)
    * Add support to **display** raw output from `LICENSE` (or similar) and `README` (or similar). (Not currently implemented in this PR.)
    * Allow users to copy a template file from theme-gem to `source_dir` with the press of a button. (Not implemented via this PR.)
  * Render a `404` error page if the site doesn't use gem-based themes.

Its a purely `GET` endpoint. I don't intend this to be used for authoring a template or edit a file within a gem.

benbalter:
Thanks for that @ashmaroli. That helps to understand what you want the API to do, but what user-facing functionality are you trying to provide? What problem are you trying to solve for? What workflow are you trying to support?

ashmaroli:
Closing in favor of #347

PR #345, opened 2017-04-12
Metadata StaticFilePicker

mertkahyaoglu:
StaticfilePicker allows users to pick a static file url from a modal that lists all of the static files for their metafield values.

Also added a tooltip to inform users which special keys they can use for a metafield in order to use `DatePicker` and `StaticfilePicker`.

Here is the demo;

staticfilepicker

### DIFF
--- a/src/containers/MetaFields.js
+++ b/src/containers/MetaFields.js
@@ -55,9 +55,15 @@ export class MetaFields extends Component {
           <a onClick={() => addField('metadata')} className="tooltip">
             <i className="fa fa-plus-circle" /> New metadata field
             <span className="tooltip-text">
-              Metadata will be stored as the <b>YAML front matter</b> within the document.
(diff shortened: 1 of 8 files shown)


src/containers/MetaFields.js
```diff
@@ -55,9 +55,15 @@ export class MetaFields extends Component {
           <a onClick={() => addField('metadata')} className="tooltip">
             <i className="fa fa-plus-circle" /> New metadata field
             <span className="tooltip-text">
-              Metadata will be stored as the <b>YAML front matter</b> within the document.
+              Metadata will be stored as the <b>YAML front matter</b> within the document.ity.
```
ashmaroli: @mertkahyaoglu **`ity.`** ? :smiley:

mertkahyaoglu: Ops 😬

mertkahyaoglu: Fixed via 2e6bb0a

PR #347, opened 2017-04-17
Front-end support for gem-based themes

ashmaroli:
(Contains diff from #344 as well)
Front-end support to easily **view** templates within theme-gems via the interface.

--
@benbalter I'm proposing a set of front-end code to **demonstrate** my intention behind #344
/cc @mertkahyaoglu


mertkahyaoglu:
Wow. That's perfect 👍 

I think it would be super useful if we can edit the theme files.

ashmaroli:
I also wanted to include a `Button` to easily copy a template from the `gem` to the `source_dir` with error messages informing / warning users if that template path already exists in the `source_dir`. But ultimately decided to leave it out for now.

Thanks for your feedback anyways, @mertkahyaoglu. Appreciate it. :smiley:

benbalter:
@ashmaroli I asked a few questions over in https://github.com/jekyll/jekyll-admin/pull/344#issuecomment-293915592:

> That helps to understand what you want the API to do, but what user-facing functionality are you trying to provide? What problem are you trying to solve for? What workflow are you trying to support?

If I'm understanding correctly, I'm a bit concerned about leapfrogging over the significantly more common use case of editing template within the site (nearly every site, but not a use case the current editor experience is optimized for), to the much less common use case of authoring a theme that exists elsewhere on the user's computer.

ashmaroli:
> I asked a few questions over in [...]

@benbalter This PR was meant to be an answer to your questions in that comment. But looks like we keep misinterpreting each other. So, I'll try to answer your questions verbally in the following comment..

ashmaroli:
> **What user-facing functionality are you trying to provide?**

Allow users **to view theme-gem contents via the admin interface**. In future, **copy a template file from theme-gem to `source_dir`** with a button-click.

--

> **What problem are you trying to solve for?**

A feature enabled by default in JekyllCore (theme-gems) is not present even in basic form via the admin interface.

--

> **What workflow are you trying to support?**

TBH, this is not part of a workflow. Rather, a feature for completion sake.
Users need not traverse multiple directories to view their theme-gem contents.(I myself have my `GEM_HOME` defined on a different drive on my Windows system.) They may do so via the admin interface with a few clicks.

PR #350, opened 2017-04-21
Add support for standard save keyboard shortcut

### LINKED ISSUE #304: Support standard keyboard shortcut to save
### Description:

Many online WYSIWYG editors such as Google Docs, Gmail, etc. hook into the standard OS save shortcut (`ctrl`+`S` on Windows, `cmd`+`S` on Mac, etc) to prevent the default browser behavior of trying to save the current page content as an html file.

This should be implemented in Jekyll admin. The current default behavior is jarring, especially because many users will be in the habit of mashing ctrl+s frequently, only to have to close the default dialogue and manually click the save button instead.

- jekyll-admin v0.3.0
- jekyll v3.3.1
- tested on Windows 10 and MacOS 10.10
- Chrome v56.0.2924.87 (64-bit)

mikefowler:
## Overview

Adds support for the standard "Save" keyboard shortcut: Command+s on OSX,
or Ctrl+s on Linux and Windows.

This change introduces a new dependency, react-hotkeys, which will make
it trivial to add additional keyboard shortcuts in the future.

Closes #304

cc: @JohannesMP @mertkahyaoglu

### DIFF
--- a/src/containers/App.js
+++ b/src/containers/App.js
@@ -1,4 +1,7 @@
 import React, { Component, PropTypes } from 'react';
+import { HotKeys } from 'react-hotkeys';
+
+import keyboardShortcuts from '../constants/keyboardShortcuts';
 
 // Components
 import Sidebar from './Sidebar';
@@ -9,7 +12,9 @@ class App extends Component {
 
   render() {
     return (
-      <div className="wrapper">
+      <HotKeys
+        keyMap={keyboardShortcuts}
+        className="wrapper">
         <Sidebar />
         <div className="container">
           <Header />
@@ -18,7 +23,7 @@ class App extends Component {
           </div>
         </div>
         <Notifications />
-      </div>
+      </HotKeys>
     );
   }
 }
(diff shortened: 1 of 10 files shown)


mertkahyaoglu:
@mikefowler This is really great 👍  

Could you add the save shortcut for `Configuration` too?

src/containers/App.js
```diff
@@ -9,7 +12,9 @@ class App extends Component {
 
   render() {
     return (
-      <div className="wrapper">
+      <HotKeys
```
mertkahyaoglu: Do we need `HotKeys` here?

mikefowler: Yes. As far as I can tell, `react-hotkeys` requires you to wrap the entire React tree that will be listening for hotkeys.

mikefowler:
@mertkahyaoglu, I added the shortcut to `Configuration` as well, and added a helper for preventing default events 👍

mertkahyaoglu:
Thank you @mikefowler , very useful addition 👍 

Great to have you here as a first-time contributor ❤️

mikefowler:
@mertkahyaoglu <3

PR #351, opened 2017-04-23
Support editing raw config file via admin interface

### LINKED ISSUE #295: Configuration component does not show comments in config file
### Description:

I'm not able to see the comments included in my `_config.yml`

* Version of JekyllAdmin I'm using <HINT: use `bundle show` to check>: `0.3.0, master`
* Version of Jekyll I'm using <HINT: use `bundle show` to check>: `3.4.0`
* Operating System <e.g. OS X, Windows>: `Windows 7`
* Browser <e.g, Safari, Chrome>: `Chrome`

### Steps to reproduce:

1. Run JekyllAdmin either via `Gemfile` or via `server-frontend` script
2. Go to `/admin/configuration`
### I expected the following:

Expected to see the comments present in my `_config.yml`

No comments
### Other details:

I feel its important that I should be able to read the included comments especially when I'm editing a `_config.yml` written by my theme-gem's developer

ashmaroli:
Fixes #295 

The interface now loads **the raw content** from the first config file found and writes to the same file.
Additional changes made to have the code for `Configuration` resemble that for `DataFiles`

- [x] Update API
- [x] Update frontend
- [x] Update docs

Thanks :smiley:

### DIFF
--- a/lib/jekyll-admin/server/configuration.rb
+++ b/lib/jekyll-admin/server/configuration.rb
@@ -2,12 +2,15 @@ module JekyllAdmin
   class Server < Sinatra::Base
     namespace "/configuration" do
       get do
-        json raw_configuration.to_liquid
+        json({
+          :content     => parsed_configuration,
+          :raw_content => raw_configuration,
+        })
       end
 
       put do
         write_file(configuration_path, configuration_body)
-        json raw_configuration.to_liquid
+        json request_payload
       end
 
       private
@@ -23,9 +26,17 @@ module JekyllAdmin
         @configuration ||= Jekyll.configuration(overrides)
       end
 
-      # Raw configuration, as it sits on disk
-      def raw_configuration
+      # Configuration data, as read by Jekyll
+      def parsed_configuration
         configuration.read_config_file(configuration_path)
+      end
+
+      # Raw configuration content, as it sits on disk
+      def raw_configuration
+        File.read(
+          configuration_path,
+          Jekyll::Utils.merged_file_read_opts(site, {})
+        )
       end
 
       # Returns the path to the *first* config file discovered
@@ -34,8 +45,12 @@ module JekyllAdmin
       end
 
       # The user's uploaded configuration for updates
+      # Instead of extracting `raw_content` directly from the `request_payload`,
+      #   assign the data to a new variable and then extract the `raw_content`
+      #   from it to circumvent CORS violation in `development` mode.
       def configuration_body
-        YAML.dump request_payload
+        payload = request_payload
+        payload["raw_content"]
       end
     end
   end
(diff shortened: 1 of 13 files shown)


src/actions/config.js
```diff
@@ -18,23 +18,33 @@ export function fetchConfig() {
...
-  return (dispatch) => {
-    let json;
-    try {
-      json = toJSON(config);
-    } catch (e) {
-      return dispatch(addNotification(getParserErrorMessage(), e.message, 'error'));
+  return (dispatch, getState) => {
+    const errors = validateConfig(config);
+    if (errors.length) {
+      return dispatch(validationError(errors));
```
mertkahyaoglu: You need to import `validationError` action to pass the errors to the store.

```javascript
import { validationError } from '../actions/utils';
```

lib/jekyll-admin/server/configuration.rb
```diff
@@ -2,12 +2,15 @@ module JekyllAdmin
   class Server < Sinatra::Base
     namespace "/configuration" do
       get do
-        json raw_configuration.to_liquid
+        json({
+          :content     => parsed_configuration,
+          :raw_content => raw_configuration,
+        })
```
mertkahyaoglu: For the tests to pass, since you put the configuration under `content` & `raw_content` fields, you need to fix the  line [here](https://github.com/jekyll/jekyll-admin/blob/master/spec/jekyll-admin/server/configuration_spec.rb#L11) (and the following ones as well if necessary) accordingly.

```ruby
expect(last_response_parsed["content"]["foo"]).to eql("bar")
```

src/actions/config.js
mertkahyaoglu: And also we need to show the errors to users if there are any. Check other views for guidance. Simply, you need to;

1.  use `Errors` component.
2. clear errors when the component will unmount.

mertkahyaoglu: And also, we need to show the errors if any to users in the Configuration view. Check other views for guidance. Simply, you need to;

1. use `Errors` component to show them.
2. and clear errors when the component unmounts.

mertkahyaoglu: Ops. Forgot to tell you about the view test. Could you please add that too? Sorry for that 😬 

In `views/tests/configuration.spec.js`;
```javascript
  it('should render error messages', () => {
    const { errors } = setup(Object.assign({}, defaultProps, {
      errors: ['The title field is required!']
    }));
    expect(errors.node).toExist();
  });
```

ashmaroli: No probs. Will add the above test too :slightly_smiling_face:

ashmaroli:
Thank you @mertkahyaoglu :clinking_glasses:

PR #354, opened 2017-04-28
Add GUI support for editing Data files

### LINKED ISSUE #297: Data file editing GUI
### Description:

Add a visual editor that allows editing, removing and moving entities without the hassle of coding YAML.

People who don't know YAML or even to program should be able to modify data simply.

Maybe something like this: http://jeremydorn.com/json-editor/ or the metadata editor of the posts
Worth mentioning Siteleaf already have an editor like this

* Version of JekyllAdmin I'm using 0.1.1
* Version of Jekyll I'm using 3.1.6
* Operating System: macOS Sierra 10.12.3
* Browser: Chrome

ashmaroli:
~~WIP~~
Fixes #297 

@mertkahyaoglu its solid enough to be tested manually via `server-frontend` 
Feedback invited.


mertkahyaoglu:
Woow! This is really cool @ashmaroli 😮

mertkahyaoglu:
I would prefer moving save, delete and toggle buttons to the right sidebar for consistency. This way we can also show dropdowns without overflowing.

ashmaroli:
> I would prefer moving save, delete and toggle buttons to the right sidebar for consistency

Yes, I'm in favor for consistency too.. just that I encountered a minor hiccup with every new `object` or `list` indenting towards the right, especially when the value set is nested 4-5 levels deep getting squashed if the `.content-body` has fixed-width or creates an ugly horizontal scrollbar when i set `overflow-x`.. guess its a compromise I'll have to make..

ashmaroli:
I'd like to ask a couple of questions regarding the `MetaArrayItem` component:
  - Is the `index + 1` really useful? Can I remove it?
  - How about scraping the entire `~array-header` className and adopting the following design:
    `[moveButton]`  .  `[listItem]`  .  `[convertButton]`

mertkahyaoglu:
> Is the index + 1 really useful? Can I remove it?

Yeah. For software people, indexes start from 0 but for normal people it should start from 1 😄  

> [moveButton] . [listItem] . [convertButton]

Could you explain this a bit more ?

ashmaroli:
> indexes start from 0 but for normal people it should start from 1

I agree but what's the point of displaying the number? It changes value on sort.

> Could you explain `[moveButton] . [listItem] . [convertButton]` a bit more ?

gui

mertkahyaoglu:
>  It changes value on sort.

Yeah, it is the purpose. Just a visual assistance to remember items order if you have so many items in your list.

mertkahyaoglu:
> [moveButton] . [listItem] . [convertButton]

Doesn't this approach make input space narrower especially for nested items ?

ashmaroli:
> Yeah, it is the purpose. Just a visual assistance to remember items order if you have so many items in your list.

Then, it should ideally be the internal `index` instead of expected `index + 1`. as then the user would be able to access the item correctly. Show the user that the third list item is actually at `data-index-2`:

```yaml
# _data/hash.yml

array:
  - entry_1
  - entry_2
  - entry_3
```
```
{{ site.data.hash.array[2] }} => "entry_3"
```

ashmaroli:
> Doesn't this approach make input space narrower especially for nested items ?

Yes, but from the demo, it looks like there's plenty of space available.

mertkahyaoglu:
> Then, it should ideally be the internal index instead of expected index + 1

I can't really follow. In which cases do you think it won't be the same as the internal index? 

```
array:
  - entry_1
  - entry_2
  - entry_3
```

Is there a possibility that the last item will not be index 3 ?

ashmaroli:
> Is there a possibility that the last item will not be index 3 ?

Yes, its a convention among all leading languages that array indices start at `0`, So the third item is always `my_array[2]`. Even if its not so in all languages, Jekyll is a Ruby app that will for sure read the `third item` as `my_array[2]`

ashmaroli:
modified color to resemble `monokai` theme in `Editor`

mertkahyaoglu:
What about `switch` button under `save` button and moving the caution message to where switch button is right now?

ashmaroli:
`toggle` button is now cast aside.
All that remains now is to update the spec files. Any suggestion for how to simulate the toggle in a `test` environment?

mertkahyaoglu:
What about GUI support for creating datafiles?

ashmaroli:
> What about GUI support for creating datafiles?

Another PR for that and another PR for `Configuration`

src/containers/DataGUI.js
```diff
@@ -0,0 +1,92 @@
+import React, { Component, PropTypes } from 'react';
```
mertkahyaoglu: Why are we duplicating this? We already have `Metafields` component that does the same thing.

src/containers/views/DataFileEdit.js
```diff
@@ -68,11 +79,42 @@ export class DataFileEdit extends Component {
...
+
+    return (
+      <div className="content-side">
+        <Button
+          onClick={this.toggleView.bind(this)}
+          type="view-toggle"
+          active={true}
+          triggered={this.state.guiView}
+          block />
+        <Splitter />
```
mertkahyaoglu: Let's remove this one.

src/containers/views/DataFileEdit.js
```diff
@@ -82,47 +124,46 @@ export class DataFileEdit extends Component {
...
-    const { path, raw_content } = datafile;
+    const { path, raw_content, content } = datafile;
     const filename = getFilenameFromPath(path);
 
     const keyboardHandlers = {
       'save': this.handleClickSave,
     };
 
+    const editorView = this.state.guiView ? " hidden" : "";
+    const uiView = this.state.guiView ? "" : " hidden";
```
mertkahyaoglu: What about rendering according to `guiView` instead of adding hidden class? For example;

```javascript
if(this.state.guiView) {
  return (
    <div id="raw-editor">...
  );
} else {
  return (
    <div id="gui">...
  )
}

File: src/containers/DataGUI.js (hunk 1 above)
Comment (ashmaroli) [inline src/containers/DataGUI.js]: Isn't that necessary to export the `class`? Not sure, though.. `Pages` have the same duplication

File: src/containers/views/DataFileEdit.js (hunk 2 above)
Comment (ashmaroli) [inline src/containers/views/DataFileEdit.js]: okay..

File: src/containers/DataGUI.js (hunk 1 above)
Comment (mertkahyaoglu) [inline src/containers/DataGUI.js]: > Isn't that necessary to export the class

I don't know what you mean by that but `metafields` component is already exported that's how we import it from other views.

> Pages have the same duplication

Code duplication and code reusing are different things. What I'm trying to say is instead of copy-pasting `MetaFields` code and calling it `DataGUI`, you can reuse the code and just import Metafields in `DataFileEdit` view.

Instead of this;
```javascript
import Metadata from '../DataGUI';
```

Do this;
```javascript
import Metadata from '../MetaFields';
```

ashmaroli: I see.. I think you commented on the wrong line. Hence the confusion.
This duplication was necessary to use a different sass partial to add styles without affecting existing styles for `Metafields`. Yes, there should be a better way..

mertkahyaoglu: Ops. I think we cannot comment the complete file. Sorry for that.

Maybe you can pass some props to add those styles and for others things to differentiate.

ashmaroli: :thinking: :thinking:

ashmaroli: There is considerable amount of difference between `MetaFields` and `DataGUI` that I think its best to keep them separate.

ashmaroli: [`fa89d19`](https://github.com/jekyll/jekyll-admin/pull/354/commits/fa89d196f9e8eca8c70aaf6ecb1a32fe13562f4a) to easily demonstrate what changed between `MetaFields` and `DataGUI`

mertkahyaoglu: Not much than we can handle 😬

ashmaroli: :smiley: You have write access to this branch, feel free to push a commit :innocent:

mertkahyaoglu: Will do 😄

ashmaroli:
Thank you @mertkahyaoglu :+1:  :smiley:

mertkahyaoglu:
@ashmaroli Could you add something to docs for this feature?

ashmaroli:
@mertkahyaoglu you missed including the `title` meta for `dataview`

ashmaroli:
> Could you add something to docs for this feature?

sure. once everything is finalized..

mertkahyaoglu:
> you missed including the title meta for dataview

If you mean `excluding`, I think it is 🆗  to leave it there. It will be undefined.

ashmaroli:
> If you mean `excluding`

Nope. I meant **`including`** :smile: 
``` javascript

before:
const { path, raw_content, ...rest } = metadata; // in DataGUI.js

now:
const { path, title, raw_content, ...rest } = metadata; // in MetaFields.js
```

mertkahyaoglu:
Yeah, `title` is included `now` 😄  I left it there because it will be undefined and won't be used. What's the problem?

ashmaroli:
the `title` we're referring to here is not the `title` of the datafile (that's undefined) but a `title` key in the data file content. Understandably, even `path` will need to be included into the `content` hash passed as `metadata` to the Component

ashmaroli:
Please refer the `template-config.yml` example datafile to understand

mertkahyaoglu:
Ah. I see now. We might have a key called `title` in a data file and if we leave it there it won't be shown to users, right?

ashmaroli:
right, so we need to `include` `title` and `path`

ashmaroli:
right so we need to **include** `title` and `path` to the `content` hash passed on the GUI

mertkahyaoglu:
Yeah, this way they are included in destructuring assignment (simply brackets) but excluded from the hash. Damn English 😬  Fixing it now.

ashmaroli:
:laughing: @mertkahyaoglu so sorry. But I think the entire data should be splat. `raw_content` can be a key too..

mertkahyaoglu:
OMG 😮  You're right. How come did I forget the code I wrote months ago? :face_palm:

mertkahyaoglu:
Thanks @ashmaroli . I think we're safe now 😬

ashmaroli:
:clap: Great Team work once again :smiley:

ashmaroli:
Encountered a bug in the GUI.

### Steps to reproduce bug
  1. Start `server-frontend`
  2. Open GUI at `datafiles/template-config.yml`
  3. Ensure data displayed is correct.
  4. Open `pages/page.md`
  5. Go to GUI for `datafiles/template-config.yml` from the current view
  6. See incorrect values for `title`, `path`, `raw_content`

mertkahyaoglu:
I'm aware of that issue. MetaSimple is not re-rendering on props change.

mertkahyaoglu:
@ashmaroli Making `MetaSimple` component [controlled](https://facebook.github.io/react/docs/forms.html#controlled-components) should properly update its value.

mertkahyaoglu:
Hmm. I can't activate the save button in GUI view now 😕

PR #357, opened 2017-04-29
Webpack Bundle Analyzer

mertkahyaoglu:
Why;

```
1. Realize what's really inside the front-end bundle
2. Find out what modules make up the most of it's size
3. Find modules that got there by mistake
4. Optimize it
```

How;
Open `report.html` in public folder generated via webpack.

screen shot 2017-04-29 at 15 41 14

**Note:** Looking at this, I realised that we should get rid of `unicode` module and use a custom slugify function instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,6 +90,7 @@
     "sass-loader": "4.1.0",
     "style-loader": "0.13.1",
     "webpack": "2.2.1",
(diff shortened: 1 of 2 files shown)


ashmaroli:
wow.. nice stuff.. :+1:

PR #360, opened 2017-05-02
Extend and lock current GUI support to just YAML and JSON data files

ashmaroli:
improves GUI for data files and tests with all Jekyll-supported data files


src/actions/datafiles.js
```diff
@@ -38,7 +38,7 @@ export function fetchDataFile(filename) {
   };
 }
 
-export function putDataFile(filename, data, source = "editor") {
+export function putDataFile(filename, data, source = "editor", ext = "") {
```
mertkahyaoglu: Why not get the extension from the filename? We have a [helper](https://github.com/jekyll/jekyll-admin/blob/master/src/utils/helpers.js#L67) for that too.

src/actions/datafiles.js
```diff
@@ -51,8 +51,11 @@ export function putDataFile(filename, data, source = "editor") {
       payload = { raw_content: data };
     } else if (source == "gui") {
       const metadata = getState().metadata.metadata;
-      const yaml_string = toYAML(metadata);
-      payload = { raw_content: yaml_string };
+      if (ext == ".yml" || ext == ".yaml") {
```
mertkahyaoglu: Let's use regex here to include uppercase ones as well.

```javascript
const yaml = /yaml|yml/i.test(ext);
```

src/components/Editor.js
```diff
@@ -21,11 +22,12 @@ class Editor extends Component {
   }
 
   render() {
-    const { content } = this.props;
+    const { content, ext } = this.props;
```
mertkahyaoglu: let's make `ext` prop `type` without `dot`.

src/components/Editor.js
```diff
@@ -21,11 +22,12 @@ class Editor extends Component {
...
+    const mode = (ext == ".json") ? "json" : "yaml";
```
mertkahyaoglu: Regex again.

src/containers/views/DataFileEdit.js
```diff
@@ -55,13 +55,14 @@ export class DataFileEdit extends Component {
...
-    const { datafileChanged, fieldChanged, putDataFile, params } = this.props;
+    const { datafile, datafileChanged, fieldChanged, putDataFile, params } = this.props;
+    const { ext } = datafile;
 
     // Prevent the default event from bubbling
     preventDefault(e);
 
     if (fieldChanged) {
-      putDataFile(params.data_file, null, 'gui');
+      putDataFile(params.data_file, null, 'gui', ext);
```
mertkahyaoglu: Again we can retrieve `ext` from `params.data_file`.

src/containers/views/DataFileEdit.js
```diff
@@ -78,9 +79,9 @@ export class DataFileEdit extends Component {
 
   renderAside() {
     const { datafile, datafileChanged, onDataFileChanged, fieldChanged, updated } = this.props;
-    const { path } = datafile;
+    const { path, ext } = datafile;
     const filename = getFilenameFromPath(path);
-
+    const guiSupport = (ext == ".yml" || ext == ".yaml" || ext == ".json") ? true : false;
```
mertkahyaoglu: Regex again.

src/containers/views/DataFileEdit.js
```diff
@@ -78,9 +79,9 @@ export class DataFileEdit extends Component {
 
   renderAside() {
     const { datafile, datafileChanged, onDataFileChanged, fieldChanged, updated } = this.props;
-    const { path } = datafile;
+    const { path, ext } = datafile;
```
mertkahyaoglu: Can be retrieved from `datafile`.

src/containers/views/DataFileEdit.js
```diff
@@ -120,7 +124,7 @@ export class DataFileEdit extends Component {
       return <h1>{getNotFoundMessage("data file")}</h1>;
     }
 
-    const { path, raw_content, content } = datafile;
+    const { path, raw_content, content, ext } = datafile;
```
mertkahyaoglu: Already have filename below this line. For this and above comments, I think it would be better to have a `getExtension` helper.

src/containers/views/DataFileEdit.js
```diff
@@ -149,12 +153,13 @@ export class DataFileEdit extends Component {
...
           }
           {
-            !this.state.guiView &&
+            !this.state.guiView && raw_content &&
               <div className="content-body">
                 <Editor
                   editorChanged={datafileChanged}
                   onEditorChange={onDataFileChanged}
                   content={raw_content}
+                  ext={ext}
```
mertkahyaoglu: `type` for this too.

spec/fixtures/site/_data/members.tsv
```diff
@@ -0,0 +1,3 @@
+name	age	sex	location
```
mertkahyaoglu: I think `csv` will do the job. Let's remove this 😄

ashmaroli: `csv` and `tsv` (supported by Jekyll 3.5 and beyond) have been included here for better futureproof

ashmaroli:
@mertkahyaoglu Thanks for the wonderful suggestions :+1:

src/utils/helpers.js
```diff
@@ -70,6 +70,16 @@ export const getFilenameFromPath = (path) => {
...
 /**
+ * Returns extension from the given path or filename
+ * @param {String} path or filename
+ * @return {String} extension
+ */
+export const getExtensionFromPath = (path) => {
+  if (!path) return '';
+  return path.substring(path.lastIndexOf('.') + 1);
+};
+
```
mertkahyaoglu: Could you add a test for this?

ashmaroli: added in [`1fcc613`](https://github.com/jekyll/jekyll-admin/pull/360/commits/1fcc613859e42a8cb47ba15de865ec65fb3091d2) .. changed the function a bit to account for paths that do not contain an extension

ashmaroli:
I'm not able to run `coveralls` or `script/cibuild-node` successfully locally on Windows. The `run` hangs at `babel-node tools/testCi.js`

src/utils/helpers.js
```diff
@@ -72,11 +72,17 @@ export const getFilenameFromPath = (path) => {
...
  * @param {String} path or filename
- * @return {String} extension
+ * @return {String} extension or nil
  */
 export const getExtensionFromPath = (path) => {
   if (!path) return '';
-  return path.substring(path.lastIndexOf('.') + 1);
+  const index = path.lastIndexOf('.');
+
+  if (index >= 0) {
```
mertkahyaoglu: Should it be `=`? I think last `.` at the beginning is not a valid path.

ashmaroli: `dotFiles` are valid filepaths.. just not in current context.

mertkahyaoglu: Yeah but since we are checking the last `.` here, dotFiles with no extension will get in the if statement. For example, `.babelrc`; It has no extension but `index` variable will be 0.

mertkahyaoglu: It would be good to add a test for dotFiles as well.

mertkahyaoglu:
Thanks @ashmaroli 👍

PR #362, opened 2017-05-04
Add GUI support to DataFileNew

ashmaroli:
- [x] add GUI to `DataFileNew`
- [x] update spec file for view
- [x] lock GUI support to just YAML and JSON files
    - ~~**Either** listen to `InputPath` value and disable `createButton`,~~
    - **or** add a form component to **`select`** the file extension.
- [x] update documentation

### DIFF
--- a/src/containers/views/DataFileNew.js
+++ b/src/containers/views/DataFileNew.js
@@ -30,7 +30,12 @@ export class DataFileNew extends Component {
 
   componentWillReceiveProps(nextProps) {
     if (this.props.updated !== nextProps.updated) {
-      const filename = this.refs.inputpath.refs.input.value;
+      let filename;
+      if (this.state.guiView) {
+        filename = this.refs.filename.value + this.refs.datatype.value;
+      } else {
+        filename = this.refs.inputpath.refs.input.value;
+      }
       browserHistory.push(`${ADMIN_PREFIX}/datafiles/${filename}`);
     }
   }
@@ -44,7 +49,7 @@ export class DataFileNew extends Component {
   }
 
   routerWillLeave(nextLocation) {
-    if (this.props.datafileChanged) {
+    if (!this.state.guiView && this.props.datafileChanged) {
       return getLeaveMessage();
     }
   }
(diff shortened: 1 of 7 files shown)


mertkahyaoglu:
There is one issue I encountered. When you change a field type in the gui mode, for example making it list, `Create` button is disabled. I think we should pass `|| fieldChanged` to create button's `active` prop.

src/containers/views/DataFileNew.js
```diff
@@ -35,57 +55,123 @@ export class DataFileNew extends Component {
...
+  }
+
+  handleClickSave(e) {
     const { datafileChanged, putDataFile } = this.props;
+
+    // Prevent the default event from bubbling
+    preventDefault(e);
+
+    let filename;
     if (datafileChanged) {
```
mertkahyaoglu: What about `fieldChanged`? Save action won't be dispatched if users directly edit in the gui mode.

ashmaroli:
> I think we should pass `|| fieldChanged` to create button's active prop.

No, that was intentional.. The `Create` button will be `activated` only if the `path` has been entered..

src/containers/views/DataFileNew.js
ashmaroli: Its an interface for `Create`. Enter the path in `InputPath` and viola!

mertkahyaoglu:
Could you try the following steps;

1. Click `New Data File`
2. Switch to GUI mode.
3. Enter filename as `test`
4. Select extension
5. Change first field type to list
6. Expect Create button to be disabled.

ashmaroli:
Uh-oh.. didn't expect that bug..
Thanks.. will update :+1:

src/containers/views/DataFileNew.js
mertkahyaoglu: Once you fix the below bug, you will face this one 😬

mertkahyaoglu: Create button is active but clicking it won't create anything.

ashmaroli: we're not sending a data object to the `put` action via the GUI, so this shouldn't matter..

ashmaroli:
I've made a minor change in the save button behavior:  It'll get `disabled` when the GUI input value for the file's basename becomes `nil`.This is required because the GUI for this view would otherwise send a request with `filename: .yml` thereby bypassing the `Errors` component

src/containers/views/DataFileNew.js
mertkahyaoglu: But I can't create the datafile 😄  I'm copy-pasting steps below;

1. Click New Data File
2. Switch to GUI mode.
3. Enter filename as test
4. Select extension
5. Change first field type to list
6. Add some array items.
7. Try to `Create` datafile.
8. Nothing happens.

ashmaroli: forgive my nearsightedness, Master.. :pray: 
pushed a patch..

mertkahyaoglu:
Awesome 👍  Thanks @ashmaroli .

PR #365, opened 2017-05-05
Don't trim spaces from metadata value-field

### LINKED ISSUE #364: Impossible to put spaces in metadata field
### Description:

Just a quick issue so that @ashmaroli can track it properly :)
It is now impossible to put space between words after the application of #338.

### Steps to reproduce:

Put a full sentence as an excerpt in a metadata field.

### I expected the following:

To be able to put a full sentence as an excerpt and submit it.

I can't put spaces between words, the space character is deleted as soon as written. It is possible to copy/paste sentences.

### Other details:

As suggested by @ashmaroli, typing between quotes is an efficient workaround :)

ashmaroli:
Fixes #364 
A regression was caused by #339 wherein the eventhandler attached to `.value-field` prevents typing a space character before the next word.

`<TextareaAutosize />` doesn't have an `onBlur` prop to dispatch trimming effectively. So simply reverting to old behavior.

### DIFF
--- a/src/components/metadata/MetaSimple.js
+++ b/src/components/metadata/MetaSimple.js
@@ -38,6 +38,11 @@ export class MetaSimple extends Component {
     updateFieldValue(nameAttr, e.target.value);
   }
 
+  handleEditableBlur(e) {
+    const { nameAttr, updateFieldValue } = this.props;
+    updateFieldValue(nameAttr, e.target.value.trim());
+  }
+
   handleDatepickerChange(date, dateStr) {
     const { nameAttr, updateFieldValue } = this.props;
     let formatted = moment(date).format("YYYY-MM-DD hh:mm:ss");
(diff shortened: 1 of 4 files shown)


mertkahyaoglu:
Of course this happens 😬  Whenver the field changes it will trim the white-spaces around it. We should trim only when saving.

ashmaroli:
> We should trim only when saving.

Yep, but that would mean, opening the entire `Object`, processing every `value` and `sub-value` before sending the payload..

mertkahyaoglu:
@ashmaroli Could you try the following for `TextareaAutosize`;

```javascript
onBlur={(e) => console.log(e.target.value.trim())}
```

ashmaroli:
> Could you try the following

@mertkahyaoglu Tried it. While the `<textarea>` HTML element correctly has the spaces trimmed, the resulting front matter meta contains those spaces. Even existing tests pass (which should fail if this *workaround* had worked)

mertkahyaoglu:
That's because trimmed value is not passed to the metadata store. We should call `updateFieldValue` action when blurred to do that. 

```javascript

onBlur={(e) => updateFieldValue(nameAttr, e.target.value.trim())}

```

But I'm not sure which of these two method is better. Calling an extra action onBlur or processing every value and sub-value before sending the payload 🤔  First looks easier.

ashmaroli:
> We should call `updateFieldValue` action

This works! Will push an update :smiley: :+1:

PR #369, opened 2017-05-10
Restore boolean data in front matter

### LINKED ISSUE #312: Metadata field can not have boolean value
Hello, i want to write unpublished post, and when i create new metadata field with name `published` and value `false`, this is what i got in the source file:

```
---
published: 'false'
---
```

`'false'` is of type `string`, so the post does not counts as unpublished and displayed in published post list.

If i change `'false'` to `false` (without quotes) and save article through jekyll-admin, this metadata field get lost.

Because of this bug, i can not work with unpublished articles. Thanks.

ashmaroli:
Fixes #312 
WIP

Front matter will now contain proper  boolean values instead of having associated string notations.
Replaced filtering out document metadata keys with `empty` values to `null` values


lib/jekyll-admin/server.rb
```diff
@@ -81,6 +81,13 @@ def namespace
       namespace = request.path_info.split("/")[1].to_s.downcase
       namespace if ROUTES.include?(namespace)
     end
+
+    # restore boolean data in front matter
+    def restored_front_matter
+      YAML.dump(front_matter).strip
+        .gsub(": 'false'", ": false")
+        .gsub(": 'true'", ": true")
```
mertkahyaoglu: Dat hack :trollface:

mertkahyaoglu:
I wonder if we can fix this in the front end.

ashmaroli:
> I wonder if we can fix this in the front end.

There may be a way out there.
The thing is only `boolean` passed via the frontend is converted to string.. existing boolean from config or the document itself is not converted.
So is `JSON.stringify()` the culprit? If so, then a Ruby hack before the data gets written would be necessary

Observation:
Pre-existing metadata:

Front end | Output on file
--- | ---
`all: true` | `all: true`
**`hidden: false`** | **`hidden: false`**

New metadata:

Front end | Output on file
--- | ---
`sold: true` | `sold: 'true'`
**`on_sale: false`** | **`on_sale: 'false'`**

mertkahyaoglu:
I think I found a front-end side solution to this.

In `MetaSimple.js`, change `handleEditableBlur` function that you added for #365  with this;

``` javascript
handleEditableBlur(e) {
    const { nameAttr, updateFieldValue } = this.props;
    let value = e.target.value;
    try {
      value = JSON.parse(value); // auto-casting
    } catch (e) {
      value = value.trim();
    }
    updateFieldValue(nameAttr, value);
  }
```

The only down-side of this `onBlur` event is that it's not fired when saved via `hot-keys` 😕

ashmaroli:
> The only down-side of this `onBlur` event is

We can probably additionally patch to activate the button after an `onBlur` event.. ? **This has the additional benefit of sending a single valuable log entry instead of continuous log messages (especially for long strings)**

Also, we'll still need to handle scenarios where the `fieldValue` is set to `null` literally (e.g. `layout: null`)

src/components/metadata/MetaArray.js
```diff
@@ -42,6 +42,7 @@ export class MetaArray extends Component {
           addField={addField}
           removeField={removeField}
           updateFieldKey={updateFieldKey}
+          changeFieldValue={changeFieldValue}
```
mertkahyaoglu: Why do we need this?

src/actions/pages.js
```diff
@@ -84,9 +84,10 @@ export function putPage(directory, filename) {
     }
     // clear errors
     dispatch({type: ActionTypes.CLEAR_ERRORS});
-    // omit raw_content, path and empty-value keys in metadata state from front_matter
+
+    // omit raw_content, path and empty 'title' value in metadata state from front_matter
     const front_matter = _.omit(metadata, (value, key, object) => {
-      return key == 'raw_content' || key == 'path' || value == '';
+      return key == 'raw_content' || key == 'path' || (key == 'title' && value == '');
```
mertkahyaoglu: Why omit only title?

lib/jekyll-admin/server.rb
```diff
@@ -59,7 +59,8 @@ def front_matter
 
     def document_body
       body = if front_matter && !front_matter.empty?
-               YAML.dump(front_matter).strip
+               YAML.dump(restored_front_matter).strip
+                 .gsub(": 'null'", ": null") # restore null values
```
mertkahyaoglu: Doesn't JSON.parse catch null values? Why do we need this?

src/actions/metadata.js
```diff
@@ -31,6 +31,14 @@ export function updateFieldKey(namePrefix, fieldKey, newKey) {
...
 }
 
+export function changeFieldValue(nameAttr, tempValue) {
+  return {
+    type: ActionTypes.CHANGE_FIELD_VALUE,
+    nameAttr,
+    tempValue
+  };
+}
+
```
mertkahyaoglu: Why are we introducing a new action ? `updateFieldValue` does the same thing.

src/actions/pages.js
ashmaroli: because declarations like `foo: false` get interpreted as a key `foo` with `no value` and hence gets stripped out from the final Front Matter

lib/jekyll-admin/server.rb
ashmaroli: Yes `JSON.parse` at front end (via your patch) catches it fine but ends up as `nil` when `JSON.parse` runs again on the `request_body` here which gets `dumped` by `YAML` as an empty value.

i.e.
JSON `{"layout": null, "foo": false}` ==> Ruby Hash `{"layout" => "nil", "foo" => false}` ==> Front Matter:
```
---
layout:
foo: false
---
```

ashmaroli: With the above patch, it becomes:
`{"layout": null, "foo": false}` ==>`{"layout" => "nil", "foo" => false}` ==> `"---\nlayout: 'null'\nfoo: false\n---\n"` ==>
```
---
layout: null
foo: false
---
```

src/actions/metadata.js
ashmaroli: `updateFieldValue` does two things: 
  - update `e.target.value`
  - set `fieldChanged: true` => enable the `Save` button

`changeFieldValue` does two things slightly differently:
  - update `e.target.value`
  - set `fieldChanged: false` => disables an already `enabled` Save` Button. (It will be re-enabled by the above action attached to an `onBlur` event.)

In other words, this new action was added to address our earlier concern regarding dispatching `save` with `<Hotkeys/>`. With this addition, a <TextareaAutosize> under change will deactivate the `Save` button until it goes out of `focus` when `updateValueField` will run to `set` the changed `value` and enable the `Save` button and allow dispatch save action

src/actions/pages.js
mertkahyaoglu: But this part `value == ''`, is checking whether the value is empty string or not, not checking whether it is falsy or truthy. So `foo: false` should be fine. Also your code will now allow empty-valued keys to not stripped out, for example `foo: ''`.

mertkahyaoglu: Actually we can do value === '' here.

src/actions/metadata.js
mertkahyaoglu: I prefer instantly saving without trying to click to outside of the input box. This is not user-friendly at all. Users will think they saved properly when they hit `ctrl+s` without knowing the save button is not activated at all because they didn't lose focus. Hotkeys should be time-saver. Let's not block it.

I started to think that we should better move all of the things that we do on blur (trimming, auto-casting) to the put actions. That seems safer and unavoidable. I can create a PR for that if you agree.

ashmaroli: > I started to think that we should better move all of the things that we do on blur (trimming, auto-casting) to the put actions.

That's a nice idea. We currently do similar for outputting a JSON data file via the DataGUI.
I'll close this after you open a new PR. :+1:

ashmaroli:
Closing in favor of #378

PR #373, opened 2017-05-13
send extracted config hash to main container components

ashmaroli:
send `config.content` to `<Header/>` and `<Sidebar/>` instead of extracting the config hash from within these components

### DIFF
--- a/src/containers/App.js
+++ b/src/containers/App.js
@@ -36,14 +36,19 @@ class App extends Component {
       <HotKeys
         keyMap={keyboardShortcuts}
         className="wrapper">
-        <Sidebar config={config} />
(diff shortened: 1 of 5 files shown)


mertkahyaoglu:
@ashmaroli Could you add a test for `App.js` whether the wrapper `div` exists based on `config.content`? So we can cover the conditional rendering which increases `Coverage` a bit more.

ashmaroli:
Since there isn't an existing spec file for `App.js`, the coverage is going to fall unless I test a considerable amount of statements..

mertkahyaoglu:
Alright 👍 . I will add a comprehensive test for that.

PR #375, opened 2017-05-14
Use single quotes in JS files

ashmaroli:
Replace all use of double-quotes with single quotes in all functional JS files following convention.

The test files have not been altered as they may contain strings that use backslashes to escape special characters or may contain an `apostrophe`


ashmaroli:
Aww snap! Apparently, ESLint expects `double-quotes` in `JSX`

src/components/Breadcrumbs.js
```diff
@@ -38,7 +38,7 @@ export default class Breadcrumbs extends Component {
     });
 
     return (
-      <ul className="breadcrumbs">
+      <ul className='breadcrumbs'>
```
mertkahyaoglu: Let's not use single quote for html attributes, only javascript.

package.json
```diff
@@ -184,7 +184,7 @@
       "no-underscore-dangle": 0,
       "no-alert": 0,
       "no-lone-blocks": 0,
-      "jsx-quotes": 1,
+      "jsx-quotes": ["error", "prefer-single"],
```
mertkahyaoglu: Let's make it `warning`.

src/components/Breadcrumbs.js
ashmaroli: fixed

PR #376, opened 2017-05-14
Style messages from `Errors`

ashmaroli:
### Demo:
errors


mertkahyaoglu:
Looks good 👍  I have a plan for inline error messaging rather than showing all at the top. But let's see when I can create a PR for that 😄  For now, this is nice.

PR #378, opened 2017-05-14
Format metadata object

mertkahyaoglu:
Replaces #369 

- Trims metadata keys and values
- Preserves YAML types


ashmaroli:
Sample data for you to test this manually while developing..
Enter below in a new `page` and `data file`

```yaml
foo: 'foo'
bar: "bar"
baz: true
lorem: false
ipsum: null
dolor: Some "random" string 'entries'
```

mertkahyaoglu:
@ashmaroli I think this one is OK. Could you patch for `null` values that you mentioned [here](https://github.com/jekyll/jekyll-admin/pull/369#discussion_r116348553) ?

Maybe @benbalter guides us to an easier solution if we missed something? 🤔

ashmaroli:
The latest commit added a regression to an earlier fix. Values with trailing spaces are not trimmed **as expected**:
  - When updating the `front matter`, the `textarea` element with trailing spaces remains as is with trailing spaces.
  - The `values` from `DataGUI` are not trimmed at all.

mertkahyaoglu:
Hmm.. strange. I'm leaving this out of release. I'm gonna tackle this a bit.

ashmaroli:
I was able to rectify locally by restoring the `onBlur` function. But either ways, its your call.. :+1:

mertkahyaoglu:
Yeah.. But `onBlur` seems like a bad solution to me. It would be better if we can fix this without users control (in put actions or in the backend). As I said earlier, `blur` event can be bypassed when hotkeys used. Solving it requires other validations which also I don't want to implement 😬

mertkahyaoglu:
Added blur back until find a better solution.

mertkahyaoglu:
I might have found the perfect solution 😬

ashmaroli:
> I might have found the perfect solution

does it cover `null` values in the front matter too?

mertkahyaoglu:
Nah, backend removes null values 😞

ashmaroli:
> Nah, backend removes null values

Thought so.. `YAML.dump` translates it to an empty entity

ashmaroli:
I've got some bad news.. its now broken for data files under GUI mode..

mertkahyaoglu:
Yeap, just fixed it.

ashmaroli:
gez all that remains is to patch the backend with a hack..

mertkahyaoglu:
Yeah, finally 😄

Issue #380, opened 2017-05-16
Create a new type of metadata field for Tags


jrsales:
### Description:
It would be excellent to have a special type of metadata field for creating posts with tags :)

### I expect the following:
Something like the labels on git, but instead of creating labels it would be fed by the site tags. That list would then allow for auto-completion when typing into an metadata field named "tag".
![screenshot from 2017-05-16 16-49-53](https://cloud.githubusercontent.com/assets/17572140/26112430/30c1c144-3a58-11e7-8942-1e83b603ea74.png)
### Other details:
I don't see the possibility to fetch the existing tags from the API neither, am I wrong ?

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider whether it can be accomplished in another way. If it cannot, please elaborate on why it is core to this project and why you feel more than 80% of users would find this beneficial.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

PR #389, opened 2017-05-21
Initialize a new advanced metadata field for tags

ashmaroli:
Inspired by #380 

  - [x] create and delete tags
      - [x] create new tags by pressing `comma` key <kbd> , </kbd>
      - [x] delete a tag by clicking on the `x`
  - [x] prevent creating duplicate tags
  - [x] also create tags with <kbd> spacebar </kbd> or <kbd> Enter </kbd>
  - [x] delete the last tag with <kbd> Backspace </kbd>
  - [x] Update metadata spec suite
  - [x] Add `autocomplete` suggestions
      - [x] Update API to enable requesting `site.tags`
      - [x] update frontend to render a list of suggestions for new tags


gsantner:
@ashmaroli Is this still a WIP or will this get discarded?

ashmaroli:
@gsantner its still in WIP.. waiting for the API to mature enough to easily facilitate `get` on the array of `site.tags`.
This won't be discarded. Something will eventually come out of the hat.. :smiley:

src/components/metadata/MetaObject.js
```diff
@@ -21,6 +21,7 @@ export class MetaObject extends Component {
       let type = 'simple';
       if (_.isObject(value)) type = 'object';
       if (_.isArray(value)) type = 'array';
+      if (_.isArray(value) && key === 'tags') type = 'simple';
```
mertkahyaoglu: Could you extract these conditions into a method which returns type given value?

src/components/metadata/MetaSimple.js
```diff
@@ -6,6 +6,7 @@ import moment from 'moment';
 import momentLocalizer from 'react-widgets-moment';
 import FilePicker from '../FilePicker';
 import 'react-widgets/dist/css/react-widgets.css';
+import MetaTags from './MetaTags';
```
mertkahyaoglu: Let's move this one line up.

ashmaroli:
This feature lacks one thing currently: there's no way to select *existing tag*.. 
So was thinking of importing the `MultiSelect` component from `react-widgets`.. :man_shrugging:

mertkahyaoglu:
That could be done in another pr maybe another release. This is quite enough IMO.

src/components/metadata/MetaObject.js
ashmaroli: Good idea. Will handle this in a separate PR..

PR #392, opened 2017-06-01
Warn user that `--watch` has been disabled

ashmaroli:
In light of numerous questions popped regarding failed auto-regeneration, at `jekyll/jekyll` and `StackOverflow`, this adds a `WARN` level logger message to inform users that auto-regeneration has been disabled and will only trigger via the admin interface.

```
 Auto-regeneration: disabled by JekyllAdmin.
                    The site will regenerate only via the Admin interface.
  JekyllAdmin mode: development
    Server address: http://127.0.0.1:4000/
```

IMO we should ship a `0.5.1` with this **a.s.a.p**

/cc @mertkahyaoglu @benbalter


mertkahyaoglu:
👍

PR #395, opened 2017-06-05
Build only commits to default branch

ashmaroli:


mertkahyaoglu:
What's the benefit of this?

ashmaroli:
Avoid unnecessary builds when anyone with write-access pushes a commit to any branch. IMO, its best that builds are triggered only when a PR is created and when the master branch gets updated..

Issue #401, opened 2017-06-20
Show current page title in browser tab, instead of "Jekyll Admin"


mariusa:
The icon is enough to identify that tab is for "Jekyll Admin".
We're editing multiple pages in multiple tabs, and having the actual page title instead of "Jekyll Admin" in browser tabs would help identify them.

Thanks

ashmaroli:
I did consider implementing this in the past but left it due to lack of clarity in what could be the best solution:

 - Should it reflect `title` or would `relative_path` be better?
 - Should it be `page['title']` or `Jekyll Admin | page['title']` or `site['title'] | page['title']`
 - Centralized implementation (how?) or modify individual `views`

mertkahyaoglu:
@ashmaroli Does it matter ?

Since Jekyll Admin runs locally which means SEO is not a think to care about, one should be enough just to distinguish tabs. Therefore, I think `page['title']` would be sufficient.

I assume the solution will be related with `react-router` as the frontend is a single page app.

ashmaroli:
Two candidates:
  - [react-document-title](https://github.com/gaearon/react-document-title)
    - Popular
    - Regularly updated
    - Declarative -- would require updating every view page
    - might be an overkill IMO
  - [redux-title](https://github.com/DJCordhose/redux-title)
    - not-so-popular
    - last commit: Oct 2016
    - supposedly syncs with redux state -- may work with our existing `metadata` reducer
    - might be enough for us.

Thoughts?

mertkahyaoglu:
👍  for `react-document-title` for the reason you listed and also for the author 💯  I will integrate this when I get a chance.

PR #402, opened 2017-06-21
Set current page title dynamically

ashmaroli:
A rudimentary resolution for #401 without external dependencies.

### DIFF
--- a/src/containers/views/Configuration.js
+++ b/src/containers/views/Configuration.js
@@ -10,7 +10,7 @@ import Button from '../../components/Button';
 import { putConfig, onEditorChange } from '../../actions/config';
 import { clearErrors } from '../../actions/utils';
 import { getLeaveMessage } from '../../constants/lang';
-import { preventDefault } from '../../utils/helpers';
+import { preventDefault, setTitleTag } from '../../utils/helpers';
 
 export class Configuration extends Component {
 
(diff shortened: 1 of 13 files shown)


src/containers/views/Configuration.js
```diff
@@ -73,6 +73,7 @@ export class Configuration extends Component {
     return (
       <HotKeys handlers={keyboardHandlers} className="single">
         {errors && errors.length > 0 && <Errors errors={errors} />}
+        {setTitleTag('Configuration')}
```
mertkahyaoglu: Any reason to insert this here ? 😬

ashmaroli: consistency FTW ?  :grin:
Didn't want `Jekyll Admin` to crop up as the tab title for no reason..

mertkahyaoglu: I mean it would be better to call it in `componentDidMount` hook.

ashmaroli: Ah.. that.. well, the thing is for `*Edit` views, I had faced errors when trying to set the title with `this.props.[entity].title` probably because the state might have been still at `isFetching`.. so decided to add it to the `render()` after `isFetching` is set to `false`.. then followed same for all other views..

mertkahyaoglu:
Closing in favor of #407

PR #406, opened 2017-06-23
Tag static files from a theme-gem with a sticker

ashmaroli:
Currently, there is no way to easily distinguish which static-file displayed in `Dropzone` is from the `site.source` and which one is from the theme-gem.

Additionally, there's no sense that users can click delete on a static-file from the theme-gem.

This PR resolves both of above.
theme-indicator


src/components/FilePreview.js
```diff
@@ -32,9 +32,18 @@ export default class FilePreview extends Component {
     }
     return (
       <div className="file-preview">
+        {
```
mertkahyaoglu: I know I've put a few inline conditions in the front-end but it's better to pull them out of return statement.

```javascript
let theme_indicator;
let btn_delete;

if(file.from_theme) {
  theme_indicator = <div...
} else {
  btn_delete = <button...
}
```

src/components/FilePreview.js
```diff
@@ -32,9 +32,18 @@ export default class FilePreview extends Component {
...
+            <span className="theme-indicator">
+              <i className="fa fa-diamond" aria-hidden="true" title="Theme Asset" />
+            </span>
+        }
         {nodeLink}
         <span className="filename">{file.path}</span>
-        <button onClick={() => this.handleClickDelete(file.path)} className="delete" title="Delete file">x</button>
+        {
+          !file.from_theme &&
+            <button onClick={() => this.handleClickDelete(file.path)} className="delete" title="Delete file">x</button>
```
mertkahyaoglu: Why should we not allow deleting theme files?

ashmaroli: > Why should we not allow deleting theme files?

While I was greatly surprised that you asked this question, on reflecting further, I'm assuming you were talking about removing the `Jekyll::StaticFile` object from the `site_payload` rather than deleting the file proper..

To answer otherwise, three things to consider:
 - The theme-gem is supposed to be a read-only object for Jekyll.
 - We have no write-access outside `site.source`.
 - When users click `delete`, on a theme-file they're sending a `delete` request to the `relative_path` of the static file which is sanitized to point to a non-existing entity in the `source_dir`. (e.g. `<theme>/assets/logo.png` will resolve as `<site.source>/assets/logo.png` which doesn't exist. Otherwise, Jekyll would've read that file in to memory instead)

There's no valid use-case to allow write-access to the theme-gem in future either

PR #410, opened 2017-06-29
Autofocus to newly created metafield

### LINKED ISSUE #400: "New metadata field" should switch focus to newly created field
After clicking on New metadata field and adding the field, please focus the cursor in the newly created field (which should be empty, not "New field 3"), so one can directly start entering the name instead of mouse clicking & selecting.

mertkahyaoglu:
When a new metafield is created, there is no action showing users that it is created. With this PR, newly created metafields are get focused.

Solution might look a bit weird to you. There is no easy way to determine which metafield is newly created because metafields are coming from an object in the redux store. So I came up with this solution;

We simply check whether a metafield key contains `New field` :trollface: .

Fixes #400

### DIFF
--- a/spec/fixtures/site/_config.yml
+++ b/spec/fixtures/site/_config.yml
@@ -1,7 +1,7 @@
 title: Your awesome title
 
 # Load the admin interface.
-gems:
+plugins:
   - jekyll-admin
 
 # Exclude these files from being processed.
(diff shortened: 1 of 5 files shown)


ashmaroli:
It works as expected! :smile: 
But since #400 is about having to replace the placeholder text manually, I think we should take a step further and have the `New field *` text _pre-selected_ on `focus()`, even on Google Chrome.. :wink:

mertkahyaoglu:
Actually we don't need to check for the key index. We can just check whether a metafield key contains `New field`.

```
const isNewField = /New field/.test(fieldKey);
```

What was I thinking 😩

PR #411, opened 2017-07-01
Release v0.6.0

mertkahyaoglu:
Recently, there are not so many major additions to the repo (actually there are but they are still in WIP and I can't find enough time to review them these days). So it's a good idea to release minor patches more often than before instead of releasing major ones once a month. This way, users will be updated more often with changes.

### DIFF
--- a/lib/jekyll-admin/version.rb
+++ b/lib/jekyll-admin/version.rb
@@ -1,3 +1,3 @@
 module JekyllAdmin
-  VERSION = "0.5.1".freeze
+  VERSION = "0.6.0".freeze
 end
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "jekyll-admin",
-  "version": "0.5.1",
+  "version": "0.6.0",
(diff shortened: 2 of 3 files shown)


ashmaroli:
> So it's a good idea to release minor patches more often than before

I agree with this, but IMO, it should be `0.6` because we have some new enhancements even if its not breathtaking.. `0.5.2` would've been fit for a release that contains only `bug-fixes`..

mertkahyaoglu:
You are right actually. I somehow started to ignore major releases and see them as minor releases. (I still don't know what will cause a major release 😬 , I guess going beyond localhost..). Early days we released lots of features with minor releases and now it feels like few enhancements does not deserve a minor release 😄  I think it's time to respect `semver` rules.

ashmaroli:
> I think it's time to respect semver rules.

Semver rules don't apply until we release a `v1.0.0` but still its best to follow the convention loosely to help our users expect things accordingly

ashmaroli:
If you get the time, do clone my fork's `master` branch. I plan to release a gem version of it in the future..

PR #421, opened 2017-10-15
Implement Ducks: Modular Redux

mertkahyaoglu:
With our current redux setup, if we want to add a new reducer, we have to add `actionTypes`, `actions` and `reducer` into separate files. This is a really tedious work. It makes more sense for these pieces to be bundled together in an isolated module.

With this PR, I migrated current structure to [ducks](https://github.com/erikras/ducks-modular-redux) model which is a modular way of creating reducers.


spec/fixtures/site/_config.yml
```diff
@@ -1,7 +1,7 @@
 title: Your awesome title
 
 # Load the admin interface.
-plugins:
+gems:
```
ashmaroli: CAUTION: This reverts a recent change..

mertkahyaoglu: You're right. But it gives me `Deprecation: The 'plugins' configuration option has been renamed to 'plugins_dir'. Please update your config file accordingly.` error. How do I fix this?

ashmaroli: That's probably because your `Gemfile.lock` is still using an older version of Jekyll. Running `bundle update` will remove the warning.

On an unrelated side-note, once `PR #322` gets merged, every other developer locked to an older version will thenceforth not see the warning..

mertkahyaoglu: Yeah that was it. `bundle update jekyll` solved the issue. Thank you 👍

PR #423, opened 2017-10-29
Babel Bump & Prop Types & Lint

mertkahyaoglu:
- Updated Babel presets to support new ES features
- Added `prop-types` module.
- Added `lint-staged` for precommit linting
- Prettier all js files under `src`
- Added document titles for drafts pages

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "precommit": "lint-staged"
   },
   "lint-staged": {
-    "*.js": ["prettier --single-quote --trailing-comma es5 --write", "git add"]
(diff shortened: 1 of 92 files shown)


ashmaroli:
Nice.. was wondering when would be a nice time to move to the `prop-types` package.. why not bump react version to `15.6` as well? (to prepare for `v16)`

mertkahyaoglu:
@ashmaroli I was planning to bump react directly to v16 in a seperate PR. There is going to be lots of changes like array returns, peer deps bumps and so on.

PR #431, opened 2017-11-21
add: ptBr.js

gkal19:
Added Brazilian Portuguese translation


mertkahyaoglu:
Thanks @gkal19 . We are working on renewing the admin panel. There will be all translations available soon.

PR #468, opened 2018-06-30
support livereload when jekyll-admin is installed

jamesluberda:
jekyll/jekyll#5142 introduced livereload browser refresh functionality to Jekyll core (as of 3.7.0), which included a change to the `start_up_webrick` method in `lib/jekyll/commands/serve.rb`. This is the method against which `jekyll-admin` applies a monkey patch to enable its functionality. However, because the monkey patch does not include the new call to start the livereload reactor, if `jekyll-admin` is installed and `jekyll serve` is passed the `--livereload` option, the server hangs on start. This PR simply adds the livereload call to the monkey patch that was added to core in jekyll/jekyll#5142 to continue to support livereload when `jekyll-admin` is installed.

Note that `script/cibuild` results in one failure, but this failure occurs even without this PR, and is unrelated to this change.

Failure in cibuild of stock jekyll-admin 0.8.0 from GitHub:

```
Running Ruby tests...
..............................................................................F............................................................................................................................

Failures:

  1) collections an individual collection entries lists directories
     Failure/Error: expect(entries.first["name"]).to eq("more posts")
     
       expected: "more posts"
            got: "test"
     
       (compared using ==)
     # ./spec/jekyll-admin/server/collection_spec.rb:94:in `block (4 levels) in <top (required)>'

Finished in 40.14 seconds (files took 5.38 seconds to load)
203 examples, 1 failure

Failed examples:

rspec ./spec/jekyll-admin/server/collection_spec.rb:90 # collections an individual collection entries lists directories
```


mertkahyaoglu:
Hi @jamesluberda . Could you fix that test pls? Sometimes order of those two files changes. Also there are some offenses complaining about trailing comma. We can ignore that rule in rubocop.yml.

jamesluberda:
Hi @mertkahyaoglu, you're very welcome, thank you! First, would you advise rolling these changes in with this PR, or should I put them in a separate one?

Second, I started by tackling the trailing commas issue with a quick update to `.rubocop.yml`:
```
Style/TrailingCommaInLiteral:
  Enabled: false
  ````
This cleared up those offenses. However I noticed that in addition to its objections to the trailing commas, rubocop seemed to have difficulty digesting some of the rules it pulls in from `jekyll` core. I assumed this is due to the fact that `jekyll-admin` is on 0.48.1, and  Jekyll core is on 0.57.2. For fun, I updated rubocop for `jekyll-admin` to 0.57.2, and found the `Style/TrailingCommaInLiteral` cop that I had just added is now obsolete, apparently (more or less) replaced by `Style/TrailingCommaInArrayLiteral` and `Style/TrailingCommaInHashLiteral`, both of which are configured in Jekyll core's `.rubocop.yml`, and which also eliminate the trailing comma offenses.

Should I just go ahead with the change that works under 0.48.1 (but halts rubocop 0.57.2), or is a rubocop upgrade in order? I admit being in new territory here, so not sure if that's an absurd question.

**Comment by mertkahyaoglu** [comment]
> would you advise rolling these changes in with this PR, or should I put them in a separate one?

You can add them to this PR 👍 

> Should I just go ahead with the change that works under 0.48.1 (but halts rubocop 0.57.2), or is a rubocop upgrade in order? 

I suggest we keep them in order with the current version of Jekyll.

**Comment by jamesluberda** [comment]
Ok, looked into resolving the test failure
```
rspec ./spec/jekyll-admin/server/collection_spec.rb:90 # collections an individual collection entries lists directories
```
In brief, it appears that the `_posts/test` folder that this test was designed against was removed in dbfa65e5, and the spec file was updated to reflect this change in 8d3e6f3c. However, the deleted folder was added back a day later in 3b8ab956, but the test remained the same, resulting in failure. Solution is either deleting the folder or changing the spec back to expect the folder; either resolves the error. Advise as to preference? I'm guessing that the re-add of the folder was an accident.

mertkahyaoglu:
Yeah, let's delete `test` folder if there are no other specs depending on it. Because we already have a nested folder `more posts` for tests.

mertkahyaoglu:
Thank you @jamesluberda ❤️

PR #481, opened 2018-11-16
allows for default content to be set in config file

dnoneill:
This PR allows for default content to be set when creating a new document. This information get set in the config file and is defined based upon collection name. i.e.

```
default_content: {
  puppies: "<div id='biography' markdown='1'>\n\n</div>\n<div id='bibliography' markdown='1'>\n\n</div>"
}
```


lexoyo:
Hi I see here that this has already been merged ?
#480

mertkahyaoglu:
Yeap. This is dupe.

mertkahyaoglu:
Actually it's closed somehow 😅

PR #486, opened 2019-03-26
The data files are getting deleted when editing in Raw Editor, fix #482

### LINKED ISSUE #482: The data files are getting deleted when editing in Raw Editor
We are using Data files in our project and when we try to edit the Data files in Raw Editor mode from the Jekyll-admin, the error shows up stating could not update file and then when we go back to Data files list the file is deleted.

When we inspect, in the console we get an error - fetch.js:53 PUT http://localhost:4000/_api/data/foobar.yml 500 (Internal Server Error)

The versions that we are using are - 
jekyll - 3.8.4
jekyll-admin - 0.8.1
OS Ubuntu 16.04 and Also on macOS 10.13.5
Browser - Chrome - 64.0.32 and 71.0.3578.98 respectively.

Any help on this matter will be highly appriciated.

lexoyo:
Following the bug report #482 

As I understand it, the problems comes from the `TextareaAutosize` component which has a `value` attribute which is supposed to be set to its inner `TextArea`'s value but is always undefined. So I had to add the line `this.refs.input.value = e.target.value;` in the change event (workaround)

Also before the value of the `TextareaAutosize` component has been change, the `value` attribute is still undefined since the workaround has not been triggered so I added the `name = this.refs.inputpath.refs.input.value || this.refs.inputpath.refs.input.props.defaultValue;`

Now it is working for me

### DIFF
--- a/spec/jekyll-admin/server/data_spec.rb
+++ b/spec/jekyll-admin/server/data_spec.rb
@@ -29,7 +29,7 @@ describe "data" do
   it "gets the index" do
     get "/data"
     expect(last_response).to be_ok
-    expect(last_response_parsed[2]).to eql(base_response)
+    expect(last_response_parsed.select { |file| file["slug"] === "data_file" }[0]).to eql(base_response)
   end
 
   it "gets an individual data file" do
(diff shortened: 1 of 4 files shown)


lexoyo:
The travis CI failed because of the order of the data files returned by a call to the `/data` route or to `JekyllAdmin::DataFile.all`

I added a filter in order to be sure to compare with the test file "data_file.yml"

mertkahyaoglu:
@lexoyo That's awesome. Thanks for the PR ❤️

PR #487, opened 2019-04-15
wip: attempt to fix base url issue, closes #471

### LINKED ISSUE #471: Jekyll baseurl not applied / Not works in subfolder (relevant for especially for reverse proxy in subfolder)
### Description:
I have set `baseurl: /wiki` in `_config.yml`  and expected to get admin available on `/wiki/admin/`, but it doesn't work. Jekyll-Admin doesn't respect the setted `baseurl`.

Especially when serving through reverse proxy this is bad, because everything in Jekyll does use baseurl, but not jekyll-admin.

* Version of JekyllAdmin I'm using 0.8.1
* Version of Jekyll I'm using 3.83
* Operating System deb9

### Steps to reproduce:
Set `baseurl: /something` and jekyll-admin listens at wrong location - `/admin` instead of `/something/admin`.

### I expected the following:
Listening at `/something/admin`.

lexoyo:
This is a work in progress, so that we can discuss it

Working
* set baseurl to `/example` in _config => api and UI served from `/example/admin` and `/example/_api`
* the web app understands the path `/example/admin/...` on the client side
* refresh works too when on a deep linked page

Not working, still to be fixes
* the fonts are not loaded because webpack does hard code the path
* if `baseurl` is not provided in the config, the server does not start
* (won't be fixed) does no support the param `--baseurl` passed to the server @kwalski [suggested here](https://github.com/jekyll/jekyll-admin/issues/471#issuecomment-420611762)

Your opinion wanted
* on the client side, the base URL is build from window.location, not from the config file since we need the URL to get the confing...
* `index.html` is served with `baseurl` injected by replacing a string, this should probably be a template? or the JS and CSS tags could by created JS after it gets the base URL
* there is no unit test for this and I don't see how to restart jekyll with different params


lexoyo:
when `baseurl` is not in `_config.yml` I initialise it to `""` here, does someone know if it is a good way to go?
https://github.com/jekyll/jekyll-admin/pull/487/files#diff-2c1334e96ff5c279bda3f589e7fbbca6R36

lib/jekyll-admin.rb
```diff
@@ -32,6 +32,9 @@ def self.site
     @site ||= begin
       site = Jekyll.sites.first
       site.future = true
+      if site.config["baseurl"].nil?
+        site.config["baseurl"] = ""
```
lexoyo: when `baseurl` is not in `_config.yml` I initialise it to `""` here, does someone know if it is a good way to go?

lexoyo:
The fonts are now loading

I suffered [3 hours for this 1 line change](https://github.com/jekyll/jekyll-admin/pull/487/commits/fbfaa4bc9dffa9ce6a15a3fa5230d505c37df3b0) :face_with_head_bandage: 

Note: the webpack config is really confusing

I'm almost done with this fix, let me know if anyone can test? I can provide instructions if your not familiar with git

mertkahyaoglu:
Do we really need to prefix `baseurl` for `_api` endpoint?

ashmaroli:
> Do we really need to prefix `baseurl` for `_api` endpoint?

If users desire to wall of the "Jekyll" aspect of their site separately from the rest of their site, I guess its better that even `/api` gets prepended by the configured `baseurl`.

ashmaroli:
IMO, if the endpoints are being prepended by the baseurl, the robust way would be to find a way to use Jekyll's `URLFilters#relative_url` instead. That way, we need not bother about misconfigured baseurl values such as `//blog`, `~blog`, and other values that will be sanitized by Jekyll for locations outside `/admin` or `/api`.

PR #492, opened 2019-08-26
Adjust gemspec to allow use with Jekyll 4.0

zocoi:
Since jejyll 4 is released, bump up the requirement version

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 _site/
 .gems/
+.jekyll-cache/
 .sass-cache/
 .jekyll-metadata
 /.bundle/
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 _site/
+.gems/
 .sass-cache/
 .jekyll-metadata
 /.bundle/
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,33 +1,20 @@
+require: 'rubocop-jekyll'
 inherit_gem:
-  jekyll: .rubocop.yml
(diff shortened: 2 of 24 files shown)


ashmaroli:
@zocoi Couple of comments (FYI):
- Dependency `"jekyll", "~> 4"` means the next release can't be run with Jekyll 3.x. That would be bad as the majority of the users would still be using Jekyll 3.x series.
Instead opt for **`"jekyll", ">= 3.3", "< 5.0"`**
- IMO, it'd be better if Travis would use the lowest common Ruby version which is `rvm 2.4`

tosbourn:
Just heard about this project and would love to see this change make it in so I can try jekyll-admin on my blog. 💜

mertkahyaoglu:
@ashmaroli I think this is ready to go, right?

ashmaroli:
> I think this is ready to go

Not quite. First, I shall merge `#497`. Then reduce the scope of changes in this PR..

mertkahyaoglu:
Good 👍  @zocoi We can merge this after conflicts are resolved then.

ashmaroli:
@zocoi I've thinned out the changes made in this branch to keep the PR focused to the primary intention.
Regarding the use of Node 8 for testing, we can bump if necessary in the future. For now, tests run on it just fine.

ashmaroli:
Thank you for your contribution @zocoi :smiley:

zocoi:
you're welcome, thanks for responding to this PR quickly :)

PR #495, opened 2019-09-23
Update installing plugins URL

ismailarilik:


ashmaroli:
Thank you @ismailarilik

PR #497, opened 2019-09-25
Use `rubocop-jekyll` to enforce Jekyll's styles

ashmaroli:
To stop inheriting `rubocop.yml` from the *latest* version of Jekyll as that may cause false CI failures for PRs in this project.

Miscellaneous developmental changes include:
- adding support for Bundler 2.0
- dropping test coverage for EOL Ruby 2.3 and older.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 _site/
+.gems/
 .sass-cache/
 .jekyll-metadata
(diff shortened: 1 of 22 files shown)


ashmaroli:
This is needed to cleanup diff at #492

PR #499, opened 2019-09-25
Update Jekyll server monkey-patch to mirror Core

ashmaroli:
## Summary

* Mark the methods as `private` since #start_up_webrick is a private method in the upstream command class
* From Jekyll 3.7 onward, local variable `server` is an instance_variable in the overridden method upstream

## Reference

The upstream method definition as of `v4.0.0`:
```ruby
    def start_up_webrick(opts, destination)
      @reload_reactor.start(opts) if opts["livereload"]

      @server = WEBrick::HTTPServer.new(webrick_opts(opts)).tap { |o| o.unmount("") }
      @server.mount(opts["baseurl"].to_s, Servlet, destination, file_handler_opts)

      Jekyll.logger.info "Server address:", server_address(@server, opts)
      launch_browser @server, opts if opts["open_url"]
      boot_or_detach @server, opts
    end
```

### DIFF
--- a/lib/jekyll/commands/serve.rb
+++ b/lib/jekyll/commands/serve.rb
@@ -2,22 +2,24 @@ module Jekyll
   module Commands
     class Serve < Command
       class << self
+        private
+
         def start_up_webrick(opts, destination)
           @reload_reactor.start(opts) if opts["livereload"]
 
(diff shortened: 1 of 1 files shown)


zocoi:
Just want to say thank you for the work to make this gem compatible with parent 4.0

PR #503, opened 2019-09-30
Collapse collection items in Sidebar

ashmaroli:
To improve UX in sites with numerous collections.
  - Renders as the first item in the Sidebar giving it ample space to expand downwards.
  - Renders a counter showing count of items within panel when collapsed.
  - Renders link to 'Posts' separately.

<table>
<tr><td>
  
</td><td>
  
</td></tr>
</table>

### DIFF
--- a/src/containers/Sidebar.js
+++ b/src/containers/Sidebar.js
@@ -62,7 +62,7 @@ export class Sidebar extends Component {
 
     return (
       <div>
-        {collectionItems.length > 0 && (
+        {collectionItems.length && (
           <li className={accordionClasses} style={{ maxHeight: panelHeight }}>
             <a onClick={this.handleClick}>
               <i className="fa fa-book" />
--- a/src/containers/Sidebar.js
+++ b/src/containers/Sidebar.js
@@ -8,23 +8,37 @@ import Splitter from '../components/Splitter';
 import { fetchCollections } from '../ducks/collections';
 import { capitalize } from '../utils/helpers';
 import { sidebar as SidebarTranslations } from '../translations';
+import classnames from 'classnames';
 import _ from 'underscore';
 
 export class Sidebar extends Component {
+  constructor(props) {
+    super(props);
+
+    this.handleClick = this.handleClick.bind(this);
+    this.state = { collapsedPanel: true };
+  }
+
   componentDidMount() {
     const { fetchCollections } = this.props;
     fetchCollections();
   }
 
+  handleClick() {
+    this.setState({
+      collapsedPanel: !this.state.collapsedPanel,
+    });
+  }
+
   renderCollections(hiddens = []) {
     const { collections } = this.props;
 
-    if (!collections.length) {
+    if (collections.length === 1 && collections[0].label === 'posts') {
       return null;
     }
 
-    return _.map(collections, (col, i) => {
-      if (_.indexOf(hiddens, col.label) == -1) {
+    const collectionItems = _.map(collections, (col, i) => {
+      if (_.indexOf(hiddens, col.label) == -1 && col.label != 'posts') {
         return (
           <li key={i}>
             <Link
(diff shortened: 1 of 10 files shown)


src/containers/Sidebar.js
```diff
@@ -8,23 +8,37 @@ import Splitter from '../components/Splitter';
...
 import { capitalize } from '../utils/helpers';
 import { sidebar as SidebarTranslations } from '../translations';
+import classnames from 'classnames';
 import _ from 'underscore';
 
 export class Sidebar extends Component {
+  constructor(props) {
+    super(props);
+
+    this.handleClick = this.handleClick.bind(this);
```
mertkahyaoglu: You don't need this. You can use arrow function for class methods.

src/containers/Sidebar.js
```diff
@@ -70,8 +122,10 @@ export class Sidebar extends Component {
 
     const defaultLinks = _.keys(defaults);
     let hiddenLinks;
+
     try {
       hiddenLinks = config.jekyll_admin.hidden_links;
+      hiddenLinks = hiddenLinks || [];
```
mertkahyaoglu: Let's move default array assignment to upper line.

src/containers/Sidebar.js
```diff
@@ -37,7 +51,45 @@ export class Sidebar extends Component {
...
+              {SidebarTranslations.collections}
+              <div className="counter">{collectionItems.length}</div>
+              <div className="chevrons">
+                <i className="fa fa-chevron-up" />
+              </div>
+            </a>
+            <ul>{collectionItems}</ul>
+          </li>
+        )}
+        {_.indexOf(hiddens, 'posts') == -1 && (
```
mertkahyaoglu: Let's use `includes` for all `indexOf`s for readability.

src/containers/Sidebar.js
```diff
@@ -37,7 +51,45 @@ export class Sidebar extends Component {
           </li>
         );
       }
+    }).filter(Boolean);
+
+    const accordionClasses = classnames({
+      'accordion-label': true,
+      collapsed: this.state.collapsedPanel,
```
mertkahyaoglu: Let's destruct `collapsedPanel` since it's used in following lines as well.

src/containers/Sidebar.js
```diff
@@ -37,7 +51,45 @@ export class Sidebar extends Component {
           </li>
         );
       }
+    }).filter(Boolean);
+
+    const accordionClasses = classnames({
+      'accordion-label': true,
```
mertkahyaoglu: You don't need to pass true for classes to be added directly. You can do this;
```js
classnames("accordion-label", { collapsed: .. })
```

src/containers/Sidebar.js
ashmaroli: I used this syntax so there'd be one less offense while migrating to webpack via #502

src/containers/Sidebar.js
ashmaroli: If I have it as:
```js
hiddenLinks = [];
hiddenLinks = config.jekyll_admin.hidden_links;
```
Then I'll have to add additional logic to test if `config.jekyll_admin.hidden_links` is `null`

src/containers/Sidebar.js
ashmaroli: Thanks. I didn't know this shorthand. :+1:

src/containers/Sidebar.js
mertkahyaoglu: Yeah, we will definitely fix that offence but not by changing our codebase. I'm gonna look into it.

src/containers/Sidebar.js
mertkahyaoglu:
```js
hiddenLinks = config.jekyll_admin.hidden_links || [];
```

ashmaroli: Of course..!! :man_facepalming:

src/containers/Sidebar.js
```diff
@@ -62,7 +62,7 @@ export class Sidebar extends Component {
 
     return (
       <div>
-        {collectionItems.length > 0 && (
+        {collectionItems.length && (
```
mertkahyaoglu: This could render `0` if there are no collections. Could you add a test for that case?

ashmaroli: Isn't `0` a falsey value in JS?
There's already a similar use in the component at `master`:
https://github.com/jekyll/jekyll-admin/blob/167b2bf3b7036aef3f88a27b44331f051ef1dbb0/src/containers/Sidebar.js#L22-L24


> Could you add a test for that case?

There's already a test for *rendering Sidebar successfully with zero collections* with a defined snapshot.
Since there's no change in the snapshot at this commit, is this still a blocker?

mertkahyaoglu: These are different cases. In this example, it returns `null` which renders nothing. But I wasn't sure `collectionItems.length` is converted to `falsy` value. If snapshot is the same, it works as expected then 👍

ashmaroli: You were right @mertkahyaoglu. This does indeed render `0`. I tested manually.

ashmaroli:
@mertkahyaoglu The diff changed slightly since your last review. Notably,
- there is no `<div />` enclosing the accordion list-item
- there won't be an empty `<div />` rendered when there are no collections + `"posts"` are hidden + drafts are not rendered.
- As a side-effect, there will now exist two `<Splitter />` if `"pages"` and `"posts"` are hidden in a site without drafts and other collections. I plan to resolve that in a separate PR.

Would be great if you could do a re-review.
Thanks.

ashmaroli:
Thank you very much, guys :smiley:

PR #506, opened 2019-10-03
Refresh styling of notification messages

ashmaroli:
Just visual changes to remove the outdated look from the notification messages..

A sample of notifications for the various levels:

key | sample
:---: | ---
**`'info'`** | info-notif
**`'success'`** | success-notif
**`'warning'`** | warning-notif
**`'error'`** | error-notif

### DIFF
--- a/src/containers/Notifications.js
+++ b/src/containers/Notifications.js
@@ -20,7 +20,59 @@ export class Notifications extends Component {
   }
 
   render() {
-    return <NotificationSystem ref="notificationSystem" />;
+    const style = {
+      Containers: {
+        DefaultStyle: {
+          padding: 15,
+          width: 360,
+          textAlign: 'center',
+        },
+      },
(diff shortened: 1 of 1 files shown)


src/containers/Notifications.js
```diff
@@ -20,7 +20,59 @@ export class Notifications extends Component {
   }
 
   render() {
-    return <NotificationSystem ref="notificationSystem" />;
+    const style = {
```
mertkahyaoglu: We don't want this here. It will be set in every render. We can move it outside the class or maybe another file.

ashmaroli: hmm.. valid point. How about making this a class property?
Would that make this static for every *instance*..?

mertkahyaoglu: Yeah, that would work.

PR #507, opened 2019-10-06
Refresh aesthetics of admin interface

ashmaroli:
* Reduce size of container headers and table cells to allow rendering more content in the viewport.
  * Replace gradients and shadows with a modern flat aethetic.
  * Reduce rounded-corners to a more subtle presence. Buttons  remain at current 4px rounded-corners.
  * Remove irrelevant mixin statements.
  * Improve staticfile-preview aesthetics.

### Before

jekyll-admin-master

static-files-master

### After

jekyll-admin-pr

static-files-pr

### DIFF
--- a/src/styles/mixins.scss
+++ b/src/styles/mixins.scss
@@ -1,32 +1,8 @@
-@mixin box-shadow($shadow...) {
-  -webkit-box-shadow: $shadow;
-     -moz-box-shadow: $shadow;
-          box-shadow: $shadow;
-}
-
-@mixin box-sizing($sizing...) {
-  -webkit-box-sizing: $sizing;
-     -moz-box-sizing: $sizing;
-          box-sizing: $sizing;
-}
-
-@mixin border-radius($radius...) {
-  -webkit-border-radius: $radius;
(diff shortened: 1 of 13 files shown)


src/styles/mixins.scss
```diff
@@ -1,32 +1,8 @@
...
-@mixin border-top-right-radius($radius...) {
-  -webkit-border-top-right-radius: $radius;
-     -moz-border-radius-topright:  $radius;
-          border-top-right-radius: $radius;
-}
+@mixin box-shadow($shadow...) { box-shadow: $shadow }
+@mixin box-sizing($sizing...) { box-sizing: $sizing }
+@mixin border-radius($radius...) { border-radius: $radius }
+@mixin border-top-left-radius($radius...) { border-top-left-radius: $radius }
+@mixin border-top-right-radius($radius...) { border-top-right-radius: $radius }
```
mertkahyaoglu: If we are going to delete these, there is no point making them mixins.

ashmaroli: Resolved by removing the mixins entirely..

PR #509, opened 2019-10-09
Freeze Ruby string literals with the pragma

ashmaroli:
`# frozen_string_literal: true` pragma at the top of a Ruby file instructs the interpreter to consider all string literals in that file as immutable or frozen.

*Additionally adds a requirement for at least Ruby 2.3.0 to allow using the `String#+@` method to obtain an unfrozen copy of a string.*

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -15,9 +15,6 @@ Metrics/BlockLength:
 Metrics/PerceivedComplexity:
   Max: 9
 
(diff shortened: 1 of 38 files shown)


ashmaroli:
> if that's the only place we are mutating

Tested various namespaces manually via the admin interface to be sure.. Didn't encounter any errors.. :grin:

PR #512, opened 2019-10-10
Expose data file's basename as attribute `name`

ashmaroli:
This is to provide `JekyllAdmin::DataFile` instances parity with primitives from Jekyll.

`Jekyll::Page`, `Jekyll::Document`, `Jekyll::StaticFile` instances exposes their *basename with extension* via Liquid and therefore the API here as either `:name` or `:basename`


ashmaroli:
> We can use this in the front-end now

Yep. That's the idea.. :smiley:

PR #514, opened 2019-10-11
Add Explorer component to render resource indexes

ashmaroli:
The front end for routes `/pages`, `/drafts` and `/data` are virtually identical except for the resource data, yet each component implements the same logic leading to repeated code.

We can extract all the repeated logic into a new React component, currently named `Explorer`.

*Note: Currently not useful to refactoring the front end for posts and other documents.*


mertkahyaoglu:
I was planning to achieve the same. I have something bigger in mind though: rewriting whole frontend from scratch with new React API because looking at the code from 3 years ago, makes me feel bad a bit. Of course, it'll take some time to do that but this could be a good start.

ashmaroli:
> I'm going to start doing it sooner or later

That's something nice to look forward to.. :+1:

src/components/Explorer.js
```diff
@@ -0,0 +1,127 @@
...
+    return items.map((entry, index) =>
+      entry.type == 'directory'
+        ? this.renderDirectoryRow(entry, type, index)
+        : this.renderFileRow(entry, type, index)
+    );
+  }
+
+  render() {
+    const { items, type, newBtnLabel, search, params } = this.props;
+    const typeLabel = type === 'datafiles' ? 'data files' : type;
```
mertkahyaoglu: We could create a constant for this later to get type labels;
```js
const TYPE_LABELS = {
  "datafiles": "data files",
  "pages": "pages",
  .. 
}
```

ashmaroli: Awesome idea!! :+1:

PR #515, opened 2019-10-14
Allow custom labels for collections in sidebar

### LINKED ISSUE #493: Feature request: Custom collection labels
The label for the Sidebar field is taken from collection name, which is not always the best option. The label then cannot contain some special characters (due to filesystem limitations).

I propose a setting in _config.yml:

```yml
jekyll_admin:
  collections:
    posts:
      label: Blog posts ✉
```

ashmaroli:
By simply setting **`sidebar_label`** key in the collection's metadata..
Do note that the provided string value will be rendered as given without any capitalization or titlization. Additionally, only the sidebar will reflect this metadata. Breadcrumbs will continue to reflect the original type.

Closes #493

### DIFF
--- a/src/containers/Sidebar.js
+++ b/src/containers/Sidebar.js
@@ -44,7 +44,7 @@ export class Sidebar extends Component {
               to={`${ADMIN_PREFIX}/collections/${col.label}`}
             >
               <i className="fa fa-book" />
-              {capitalize(col.label)}
+              {col.sidebar_label || capitalize(col.label)}
             </Link>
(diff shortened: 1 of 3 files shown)


mertkahyaoglu:
@ashmaroli I forgot to mention doc about this feature. Where should we mention this?

ashmaroli:
I intentionally left it undocumented (for now) so that new users don't think this feature is available in the latest release `v0.9.0`. I shall include it in the README when we near the next release date.

PR #517, opened 2019-10-16
Use Node's 'path' to compute basename and extname

ashmaroli:
Simplifies the implementation logic.

@mertkahyaoglu Was there a reason why you choose to go with the current implementation instead of using the `'path'` package?


mertkahyaoglu (approved):

mertkahyaoglu:
@ashmaroli Nope, I wasn't aware of it by that time 😅

PR #520, opened 2019-10-18
Filter away virtual pages that output to HTML file

### LINKED ISSUE #428: An inconsistency due to `jekyll-redirect-from`
Hello there !

### Description:

I've just noticed that if you use `jekyll-admin` whereas `jekyll-redirect-from` is already present and set, you'd get a "fake" entry within the **Pages** section :

image

Of course, as it's only a redirection, this page is empty :

image

* Version of JekyllAdmin I'm using : 0.7.0
* Version of Jekyll I'm using : 3.6.2
* Version of NodeJS I'm using : 8.9.1
* Operating System : Debian (Stretch)
* Browser : Firefox 57

### Steps to reproduce:

1. Install `jekyll-redirect-from` plugin
2. Use it on a page / post (a redirecting-only HTML page will be created directly into `_site/` folder)
3. Install `jekyll-admin`, and go to its **Pages** section
4. You should get a fake entry as described above

### I expected the following:

Not having this "fake" page within this section, because it's a plugin-based page !

---

Thanks (for everything),
Bye :wave:

ashmaroli:
There is no point in rendering routes to *virtual pages*&mdash;those that do not have a physical presence within the `source` directory. Such `page` objects are always generated via a plugin.

For example, `jekyll-redirect-from` plugin generates a *virtual `page`* named `redirect.html`. Such pages exist only in-memory at runtime.

Fixes #428

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,6 +8,9 @@ gemspec
 # To allow testing with specific Jekyll versions
 gem "jekyll", ENV["JEKYLL_VERSION"] if ENV["JEKYLL_VERSION"]
 
+# Fixture site dependencies
+gem "jekyll-redirect-from"
+
 # Site dependencies
 gem "jekyll-seo-tag"
 gem "jekyll-sitemap"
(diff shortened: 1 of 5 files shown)


ashmaroli:
That was fast :smiley:

PR #524, opened 2019-10-23
Style metafields similar to DataGUI styles

ashmaroli:
So that the section takes up lesser vertical space and additionally mirrors how users would edit front matter on text-editors.

### on `master`

on master

### proposed look

proposed look

---

Additionally, has modifications to `datagui.scss` to remove repeating style definitions


mertkahyaoglu:
Looks really good 👍

PR #532, opened 2019-11-15
Add Accordion component

ashmaroli:
This is primarily to optimize the `Sidebar` component.
Currently, Sidebar is rendered fresh every time the `"Collection"` list-item accordion is opened and closed. Extracting into a separate component allows Sidebar to be rendered just once.


src/components/CounterAccordion.js
```diff
@@ -0,0 +1,51 @@
...
+    const accordionClasses = classnames('accordion-label', { collapsed });
+    const panelHeight = collapsed ? minHeight : (count + 1) * minHeight;
+
+    return (
+      <li
+        className={accordionClasses}
+        style={{ maxHeight: panelHeight }}
+        onClick={this.handleClick}
+      >
+        <a>
```
mertkahyaoglu: We can make this `div` now or remove completely.

ashmaroli: I agree.. but I stuck with `li > a` structure in order to avoid adding more rules to `sidebar.scss`.
In the present form, the elements will simply use existing styles..

Whatsay?

ashmaroli:
The component has been rewritten to remove `<li>` and `<a>` tags entirely..

src/components/Accordion.js
```diff
@@ -0,0 +1,68 @@
...
+        <div className="chevrons">
+          <Icon name="chevron-down" />
+        </div>
+      );
+    }
+    return (
+      <div className="chevrons">
+        <Icon name="chevron-up" />
+      </div>
+    );
```
mertkahyaoglu:
```js
if (collapsed && counter) {
  return <div className="counter">{count}</div>;
}
return (
  <div className="chevrons">
    <Icon name={`chevron-${collapsed ? "down" : "up"}`} />
  </div>
);
```

ashmaroli: Wow.. I had a hunch this could be DRYed up further but just couldn't place it.
I love this..!!

ashmaroli:
TODO: Currently, the parent component calculates the no. of items within the accordion and passes that down as props. Ideally, the `count` should be handled by `<Accordion />` itself..

PR #534, opened 2019-11-18
Allow publishing drafts via the interface

ashmaroli:
Publish a draft with a couple of clicks..

### DIFF
--- a/src/containers/views/DraftEdit.js
+++ b/src/containers/views/DraftEdit.js
@@ -4,6 +4,7 @@ import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
 import { browserHistory, withRouter } from 'react-router';
 import _ from 'underscore';
+import moment from 'moment';
 import DocumentTitle from 'react-document-title';
 import { HotKeys } from 'react-hotkeys';
 import Button from '../../components/Button';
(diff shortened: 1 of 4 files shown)


src/containers/views/DraftEdit.js
```diff
@@ -157,6 +195,13 @@ export class DraftEdit extends Component {
                 block
               />
               <Splitter />
+              <Button
+                onClick={() => this.handleClickPublish(path)}
+                type="publish"
+                icon="send-o"
+                active={true}
```
mertkahyaoglu: No need for `true`. Just pass `active`.

src/ducks/drafts.js
```diff
@@ -106,6 +111,24 @@ export const deleteDraft = (directory, filename) => dispatch => {
     );
 };
 
+export const publishDraft = (directory, filename) => (dispatch, getState) => {
+  const metadata = getState().metadata.metadata;
+  let { raw_content, title } = metadata;
```
mertkahyaoglu: This isn't changing. We can make it `const`?

Also, `title` is not used?

src/containers/views/DraftEdit.js
ashmaroli: This was in sync with existing JSX on `master`:
https://github.com/jekyll/jekyll-admin/blob/a592eafc18611eae46a0182a9cae9f203854b69e/src/containers/views/DraftEdit.js#L152-L166

I'll send a PR to fix `master` first..

src/ducks/drafts.js
ashmaroli: Oops.. Good catch. :+1:

ashmaroli:
As per current implementation, a draft's title won't be used to determine the generated post's filename.
Only the draft's basename (with extension) is considered.

We *may* use the draft's title to determine generated post's title in a future release.

PR #535, opened 2019-11-20
Move common function in ducks to utility module

ashmaroli:
At the time the filter function was written for data files, the latter resources didn't have a `name` property. However, now they do.
But on modifying the function to filter by filename, it became evident that the functions defined in ducks for pages, drafts and data files **are now identical** and can therefore be extracted as a utility function.


ashmaroli:
Thank you.

PR #542, opened 2019-11-21
Compact switch statement in Button component

ashmaroli:
Instead of using separate statements involving dot-notation, *bracket-notation* can be used to dynamically interpolate property key and consequently reduce the no. of statements in the `switch()` expression.

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -48,7 +48,7 @@ export default function Button({
     case 'delete':
     case 'upload':
     case 'publish':
-      label = labels[type]['label'];
+      label = labels[type].label;
       break;
     default:
   }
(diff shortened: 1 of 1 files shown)


src/components/Button.js
```diff
@@ -27,28 +27,19 @@ export default function Button({
...
-      break;
-    case 'delete':
-      label = labels.delete.label;
-      break;
-    case 'view':
-      label = labels.view.label;
-      break;
-    case 'upload':
-      label = labels.upload.label;
+      label = labels[type]['label'];
```
mertkahyaoglu: No need to use brackets for literals.
```suggestion
      label = labels[type].label;
```

ashmaroli: Yes, that is evident from the deleted lines.
I chose to use bracket-notation with string literals to avoid mixing the two syntaxes together.

I'll make the changes..

PR #543, opened 2019-11-22
Import site metadata into admin interface

ashmaroli:
To allow selecting existing categories, layouts and tags via the interface.

The idea is to fetch the *data* related to above every time a page / draft / document is created / edited so that the `Metafields` component always reflects the latest status (i.e. a category added to one post will be available for the next post being edited).

Current;y, this is just Phase I of supporting categories and tags via the interface.
Phase II will involve `Metafields` utilizing the imported site meta.


ashmaroli:
> I think we can fetch site meta once in App.js and use it in other places.

Fetching just once is not enough.

The reason we need to continuously fetch is because the data exposed at this endpoint (`categories` and `tags`) can change per page / document and affect the entire site
i.e. when `site.tags` (at backend) is initially `{}`, saving a metafield `tags` with value `["foo", "bar"]` at `/admin/collections/posts/2019-12-03-hello.md` will cause the backend object `site.tags` to be populated with those keys.
The user would then expect to be suggested to use `"foo"` as they start to type `"fo"` in the `tags` metafield at location `/admin/collections/posts/2019-12-03-hello-again.md`. For that the frontend `state.meta.site.tags` should've already updated in sync with the backend.

Similarly, a layout created outside the admin interface (with the interface already loaded in the browser) should be reflected in the metafield data once the user mounts an `*Edit` or `*New` components.


> destructed in several places but not used

:facepalm: I guess I missed it when I was adding changes to git index. All those places will contain a call to `fetchSiteMeta()`

ashmaroli:
I think I can centralize the `fetchMeta()` calls and yet update metadata as required by having  it be called inside the `Metafields` container...

PR #546, opened 2019-11-27
Convert DataGUI into a bona fide component

ashmaroli:
In order to minimize repeated logic.
(However, there's still some logic repeated in the three components that consume `DataGUI`: `DataFileNew`, `DataFileEdit` and `Configuration`. But that seems to be necessary at the moment.)

*Currently on `master`, `DataGUI` is just `Metafields` rendered with prop `dataview: true` and the inputs for editing data file path being additionally implemented for both `DataFileNew` and `DataFileEdit`.*

Another internal change is to use input elements' `ref` value on save instead of setting the parent component's `state` and then reading from that value.


src/containers/views/Configuration.js
```diff
@@ -51,16 +53,29 @@ export class Configuration extends Component {
...
-    } = this.props;
+  renderContentBody() {
+    const { config, editorChanged, onEditorChange } = this.props;
     const { raw_content, content } = config;
+
+    if (this.state.guiView && content) {
+      return (
+        <DataGUI source={{ content }} onChange={onEditorChange} restricted />
+      );
+    } else {
```
mertkahyaoglu: No need to write `else` if we are returning in `if`.

src/containers/views/DataFileNew.js
```diff
@@ -78,57 +69,69 @@ export class DataFileNew extends Component {
...
   }
 
+  renderContentBody() {
+    const { datafileChanged, onDataFileChanged } = this.props;
+
+    if (this.state.guiView) {
+      return (
+        <div className="content-body">
+          <DataGUI
+            source={{ content: {} }}
```
mertkahyaoglu: We can define `defaultProps` for this in `DataGUI`.

ashmaroli: Resolved..

PR #548, opened 2019-11-29
Add StaticMetaFields component

ashmaroli:
Instead of *injecting* **default front matter** into the page / document on hitting the `save` button, simply *display* that metadata as non-editable fields above the container for editable metadata fields.

`StaticMetaFields` are entirely *static*. They can't be interacted with, and is *not connected* with the interface's *redux store*.
To override a default front matter value, users will have to manually `Add a new metafield`.

## Preview

front matter defaults


src/components/metadata/StaticMetafields.js
```diff
@@ -0,0 +1,37 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import _ from 'underscore';
+import StaticMetaField from './statics/StaticMetaField';
+import Icon from '../Icon';
+
+export class StaticMetaFields extends Component {
```
mertkahyaoglu: These all can be a function component since they won't have lifecycle methods

src/components/metadata/statics/StaticMetaSimple.js
```diff
@@ -0,0 +1,24 @@
...
+import TextareaAutosize from 'react-textarea-autosize';
+
+export class StaticMetaSimple extends Component {
+  render() {
+    const { fieldValue } = this.props;
+    return (
+      <div className="meta-value">
+        <TextareaAutosize
+          className="field value-field"
+          value={`${fieldValue}`}
```
mertkahyaoglu:
```suggestion
          value={fieldValue}
```

src/components/metadata/StaticMetafields.js
ashmaroli: Of course..!! :man_facepalming: 
I forgot about that..

src/components/metadata/statics/StaticMetaSimple.js
ashmaroli: This is intentional.. The same also exists in the *non-static* version `MetaSimple`:
https://github.com/jekyll/jekyll-admin/blob/cbc5ecc90b7aaf1aee76b517793afd434c99f901/src/components/metadata/MetaSimple.js#L51

I'm guessing that this was to ensure that even Booleans are stringified..

mertkahyaoglu: Hmm.. You're right. Leave it then 👍

ashmaroli:
@mertkahyaoglu Updated class components to be function components. Snapshots from the initial commit remains unchanged..

Requesting a re-review before merging..

mertkahyaoglu:
LGTM 👍

PR #554, opened 2019-12-08
Add categories, tags, layouts to fixture site

ashmaroli:
To allow testing their utilization via the interface.
Also adds a couple of static files to the fixture collection `"puppies"`


ashmaroli:
To allow testing their utilization via the interface.
Also adds a couple of static files to the fixture collection "puppies"

PR #557, opened 2019-12-08
Render `<button>` when href attr will be undefined

ashmaroli:
`<a>` tag with an undefined `href` attribute is not a link per se and therefore the `target` attribute on such an element is redundant as well.
In such circumstances, `<Button/>` should generate an HTML `<button>` element.

With this change `Button` component requires either `to` or an `onClick` prop to render an element.


src/components/Button.js
```diff
@@ -51,22 +51,27 @@ export default function Button({
...
+    return (
+      <a href={to} target="_blank" className={btnClass}>
+        {iconName && <Icon name={iconName} />}
+        {label}
+      </a>
+    );
+  } else if (onClick) {
+    return (
+      <button onClick={onClick} className={btnClass}>
+        {iconName && <Icon name={iconName} />}
```
mertkahyaoglu: We can extract this into a variable and reuse it in both places.

ashmaroli:
TODO: Forgot to update styles..

PR #560, opened 2019-12-12
Extract static-file picker logic into a component

ashmaroli:
In order to:
  - remove modal state from `MetaSimple` component.
  - facilitate wiring into MarkdownEditor in future.

### DIFF
--- a/src/components/metadata/MetaSimple.js
+++ b/src/components/metadata/MetaSimple.js
@@ -7,24 +7,12 @@ import moment from 'moment';
 import momentLocalizer from 'react-widgets/lib/localizers/moment';
 import StaticIndex from '../../containers/views/StaticIndex';
 import Icon from '../Icon';
+import FilePicker from '../FilePicker';
 import 'react-widgets/dist/css/react-widgets.css';
 
 momentLocalizer(moment);
 
 export class MetaSimple extends Component {
-  state = {
-    staticfiles: [],
-    showModal: false,
-  };
-
-  handleOpenModal = () => {
-    this.setState({ showModal: true });
-  };
-
-  handleCloseModal = () => {
-    this.setState({ showModal: false });
-  };
-
   handleEditableChange = e => {
     const { nameAttr, updateFieldValue } = this.props;
     updateFieldValue(nameAttr, e.target.value);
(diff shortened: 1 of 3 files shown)


src/components/metadata/MetaSimple.js
```diff
@@ -7,24 +7,12 @@ import moment from 'moment';
...
 import StaticIndex from '../../containers/views/StaticIndex';
 import Icon from '../Icon';
+import FilePicker from '../FilePicker';
 import 'react-widgets/dist/css/react-widgets.css';
 
 momentLocalizer(moment);
 
 export class MetaSimple extends Component {
-  state = {
-    staticfiles: [],
```
mertkahyaoglu: Why were using this and not using it now?

src/components/metadata/MetaSimple.js
```diff
@@ -83,36 +70,7 @@ export class MetaSimple extends Component {
           value={fieldValue}
           ref="imagepicker"
         />
-        <span className="images-wrapper">
-          <button onClick={this.handleOpenModal}>
-            <Icon name="picture-o" />
-          </button>
-          <Modal
-            isOpen={this.state.showModal}
-            onAfterOpen={this.fetchStaticFiles}
```
mertkahyaoglu: Not fetching static files now?

src/components/metadata/MetaSimple.js
ashmaroli: It was a development artifact. If you inspect the [commit](https://github.com/jekyll/jekyll-admin/commit/9531f8966e3a2cd4a6299a07cc3ed4dbf0d666d0) that introduced this, and [the file directly in the final unified diff](https://github.com/jekyll/jekyll-admin/blob/ead9cffc087bf135e9db17e3ea2be81564042bf8/src/components/metadata/MetaSimple.js), you'll see that `{this.state.staticfiles}` and `{this.fetchStaticFiles}` had already become redundant.

src/components/metadata/MetaSimple.js
ashmaroli: As noted in https://github.com/jekyll/jekyll-admin/pull/560#discussion_r357512404, this is a development artifact.
To clarify, the child element of the Modal, `StaticIndex` handles the fetch..

PR #563, opened 2019-12-18
Mock `window.confirm()` in container-view tests

ashmaroli:
So that `actions.delete<type>` is tested properly.

### DIFF
--- a/src/containers/views/tests/datafileedit.spec.js
+++ b/src/containers/views/tests/datafileedit.spec.js
@@ -84,8 +84,9 @@ describe('Containers::DataFileEdit', () => {
 
   it('should call deleteDataFile', () => {
     const { deleteButton, actions } = setup();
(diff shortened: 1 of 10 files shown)


src/containers/views/tests/datafileedit.spec.js
```diff
@@ -84,8 +84,9 @@ describe('Containers::DataFileEdit', () => {
 
   it('should call deleteDataFile', () => {
     const { deleteButton, actions } = setup();
+    window.confirm = jest.fn(() => true);
```
mertkahyaoglu: I think we can mock this globally in `tools/fileMock.js`.

ashmaroli: I have moved the mocks into `tools/fileMock.js` and renamed it into `tools/jestMocks.js`...

PR #568, opened 2019-12-23
Reduce syntax-highlighting support in MDEditor

ashmaroli:
In order to avoid bloating the production `bundle.js` with rarely used languages.


ashmaroli:
### Bundlesize stats

branch | `bundle.js` | `bundle.js.map`
--- | --- | ---
`master` | 2.24MB | 6.74MB
`PR` | 1.76MB | 6.02MB

src/utils/highlighter.js
```diff
@@ -0,0 +1,194 @@
...
+hljs.registerLanguage('vbnet', require('highlight.js/lib/languages/vbnet'));
+hljs.registerLanguage('vbscript', require('highlight.js/lib/languages/vbscript'));
+hljs.registerLanguage('vbscript-html', require('highlight.js/lib/languages/vbscript-html'));
+// hljs.registerLanguage('verilog', require('highlight.js/lib/languages/verilog'));
+// hljs.registerLanguage('vhdl', require('highlight.js/lib/languages/vhdl'));
+hljs.registerLanguage('vim', require('highlight.js/lib/languages/vim'));
+// hljs.registerLanguage('x86asm', require('highlight.js/lib/languages/x86asm'));
+// hljs.registerLanguage('xl', require('highlight.js/lib/languages/xl'));
+// hljs.registerLanguage('xquery', require('highlight.js/lib/languages/xquery'));
+hljs.registerLanguage('zephir', require('highlight.js/lib/languages/zephir'));
```
mertkahyaoglu: I think we can store keys of languages we want to support in an array constant then register them in a loop.
It would be easier to see which languages are enabled that way.

ashmaroli: Technically, we can. But webpack won't evaluate the `require` code then and consequently include all languages in the bundle. To avoid that, we'll have to configure webpack to remove the unneeded languages..

mertkahyaoglu: OK. Never mind then.

ashmaroli: Done.

ashmaroli:
### Bundlesize stats

branch | `bundle.js` | `bundle.js.map`
--- | --- | ---
`master` | 2.22MB | 6.73MB
`PR` | 1.75MB | 6.00MB

PR #572, opened 2019-12-27
Refactor ducks

ashmaroli:


src/ducks/metadata.js
```diff
@@ -62,20 +57,9 @@ export const convertField = (nameAttr, convertType) => ({
...
-export const updateBody = body => ({
-  type: UPDATE_BODY,
-  body,
-});
-
-export const updatePath = path => ({
-  type: UPDATE_PATH,
-  path,
-});
+export const updateTitle = title => ({ type: UPDATE_TITLE, title });
```
mertkahyaoglu: One-line is not always good. I had the same feeling years ago but today I find them unnecessary. They are not only hard to read but also hard to make modifications. For example, if we need to do a calculation inside this function, we have to add paranthesis and return statement at the end. Another think is we don't have any concern on too many LOCs or big filesizes. All of the code is minified at the end.

src/ducks/config.js
```diff
@@ -42,9 +42,7 @@ export const putConfig = (config, source = 'editor') => (
 
   // handle errors
   const errors = validateConfig(config);
-  if (errors.length) {
-    return dispatch(validationError(errors));
-  }
+  if (errors.length) return dispatch(validationError(errors));
```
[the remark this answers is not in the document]
ashmaroli: okay..

src/ducks/metadata.js
ashmaroli: okay..

ashmaroli: > All of the code is minified at the end.

True. But the sourcemap isn't minified. That's what increases the size of the published gem.

mertkahyaoglu: Hmm. We shouldn't need sourcemap for production though.

ashmaroli: That is up for debate. However, the following might help:
https://css-tricks.com/should-i-use-source-maps-in-production/

src/ducks/collections.js
```diff
@@ -145,34 +126,40 @@ export const putDocument = (collection, directory, filename) => (
...
-    }
-    return `${date}-${slugify(metadata.title)}.md`;
+    const docDate = date ? date.split(' ')[0] : moment().format('YYYY-MM-DD');
+    return `${docDate}-${slugifiedTitle}.md`;
   }
-  return `${slugify(metadata.title)}.md`;
+  return `${slugifiedTitle}.md`;
+};
+
+const validateMetadata = (metadata, collection, directory) => {
```
mertkahyaoglu: This is not just validating metadata but also mutating `path`. I don't think we should mix things here. In order to understand what this function does to path, we need to look into it. It seems not write to me.

ashmaroli: So, what would be the best way to proceed yet avoid repeating the following logic in every duck (that handles `metadata`)?
```js
const validateMetadata = (metadata, directory) => {
  let { path, title } = metadata;
  let errors = [];

  if ((!path || `${path}/` === directory) && title) {
    path = getFilenameFromTitle(title);
  } else {
    errors = validatePage(metadata);
  }
  return { path, errors };
};
```

mertkahyaoglu:
@ashmaroli Is this ready to go?

PR #573, opened 2019-12-28
Integrate react-scripts

mertkahyaoglu:


ashmaroli:
Couple of stuffs:
  - Why is `node-sass` a dependency?
  - I don't think we should have `alt="Jekyll Admin"` for all images in `FilePreview` :smile: 
  - Can the following be defined in `setupTests.js`?
    ```js
    .reply(200, null, {
      'Access-Control-Allow-Origin': '*',
      'Content-Type': 'application:json',
    }
    ```
  - Bundlesize is failing on AppVeyor.. Increased size due to `\r\n` perhaps?

mertkahyaoglu:
@ashmaroli 

> Why is node-sass a dependency?

react-scripts expects that module for sass files. https://create-react-app.dev/docs/adding-a-sass-stylesheet/

> I don't think we should have alt="Jekyll Admin" for all images in FilePreview

Actually it doesn't really matter for local. Just put it there to get rid of the eslint rule. I might even disable that rule.

> Can the following be defined in setupTests.js?  

I don't think so.

ashmaroli:
> react-scripts expects that module for sass files

Local tests show that it doesn't matter if `node-sass` is a devDependency. The css files are split into chunks either ways.

I couldn't test the production build since the `index.html` references assets without prepending `/admin/`. So, this can't be merged in the current state.

> Actually it doesn't really matter for local.

Since we're adding an alt-text, I think `alt={file.relative_path}` would be better..

ashmaroli:
@mertkahyaoglu Thank you for revisiting this PR. Would like your inputs on the following:
- When I ran `yarn install`, node-sass-4.13.1 got installed. Should I update the `yarn.lock` in this branch?
- When I run `yarn start`, the browser opens `http://localhost:3000` instead of `http://localhost:3000/admin`. Is that the same way at your end?
- When I run `yarn test`
  - The test-suite runs under *watch-mode* by default. Can the *watch-mode* be opt-in instead?
  - The tests fail because the *generated* snapshots for `<Header/>` has version rendered as `0.10.0`..
- Lastly, `script/cibuild-node` fails locally because `test:CI` is no longer a valid npm script.

mertkahyaoglu:
@ashmaroli Thanks for the review!

Fixed all except opening the app at `/admin` path. Couldn't figure out how to change the config for that. Will investigate further.

mertkahyaoglu:
@ashmaroli Looks like it was a bug from `react-scripts` side. I updated it to [3.4.0](https://github.com/facebook/create-react-app/releases/tag/v3.4.0) and they fixed setting `PUBLIC_URL` with that version. Now it opens browser in `/admin` path 👍

ashmaroli:
@mertkahyaoglu Thank you for digging the fix out.
- [x] Development build at `http://localhost:3000/admin/` - **Success**
- [ ] Production build at `http://localhost:4000/admin/`  - **Failed**

Is the production URL still `http://localhost:4000/admin/`?

PR #576, opened 2019-12-30
Use Yarn to manage dependencies on CI

ashmaroli:
Since we already check-in `yarn.lock` to the repository, it would be better if CI builds used the same set.

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -4,4 +4,4 @@ set -e
 
 script/branding
 bundle install
-npm install
+yarn install


script/bootstrap
```diff
@@ -4,4 +4,4 @@ set -e
 
 script/branding
 bundle install
-npm install
+yarn install
```
mertkahyaoglu: FYI: just `yarn` was enough 😄 It can stay like this though.

PR #578, opened 2020-01-09
Use fetch wrapper for DELETE requests to dispatch

ashmaroli:
Currently, a `fetch` wrapper for `DELETE` requests, is left unused.
Using the wrapper with slight modifications will bring consistency with `GET` and `PUT` requests that are sent via respective wrappers.

### DIFF
--- a/src/utils/fetch.js
+++ b/src/utils/fetch.js
@@ -93,12 +93,12 @@ export const del = (url, action_success, action_failure, dispatch) => {
     method: 'DELETE',
     credentials: 'same-origin',
   })
-    .then(data => {
+    .then(data =>
       dispatch({
         type: action_success.type,
         id: action_success.id,
-      });
-    })
+      })
+    )
     .catch(error => {
       dispatch({
         type: action_failure.type,
--- a/src/utils/fetch.js
+++ b/src/utils/fetch.js
@@ -7,7 +7,7 @@ const {
   getErrorMessage,
   getFetchErrorMessage,
   getUpdateErrorMessage,
-  getDeleteMessage,
+  getDeleteErrorMessage,
 } = translations;
 
 /**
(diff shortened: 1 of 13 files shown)


src/utils/fetch.js
```diff
@@ -93,12 +93,10 @@ export const del = (url, action_success, action_failure, dispatch) => {
...
   })
-    .then(data =>
-      dispatch({
-        type: action_success.type,
-        id: action_success.id,
-      })
-    )
+    .then(data => {
+      dispatch({ type: action_success.type });
+      dispatch(action_success.update);
```
mertkahyaoglu: Why are there 2 dispatch here?

ashmaroli: @mertkahyaoglu  Because the original logic sends two dispatches. For example:
```js
export const deleteDraft = (directory, filename) => dispatch => {
  return fetch(draftAPIUrl(directory, filename), {
    method: 'DELETE',
    credentials: 'same-origin',
  })
    .then(data => {
      dispatch({ type: DELETE_DRAFT_SUCCESS });
      dispatch(fetchDrafts(directory));
    })
    .catch(error =>
      dispatch({
        type: DELETE_DRAFT_FAILURE,
        error,
      })
    );
```

mertkahyaoglu: Oh, is it because we are not updating state, just refetching all drafts again, right? That's bad actually 😄

ashmaroli: Hmm.. this double dispatch could also be the reason behind the race condition issues with the frontend layer and the Ruby backend.. not sure though.

ashmaroli: Fixed this double-dispatch in https://github.com/jekyll/jekyll-admin/pull/578/commits/b11903bf670549a540cd38faa6281b76a2f4482e

PR #580, opened 2020-01-14
Feature Request: Make the use of SimpleMDE optional

cs:
In certain situations, for instance, when you want to use the Google Chrome plugin from Grammarly.com, the use of editors based on `contenteditable` is not supported. Therefore, I'd like Jekyll Admin to offer the possibility to fall back to plain `<textarea>`s. I see two ways to implement this:

1. Make it a configuration option (i.e., `_config.yml`) where the user can pick between SimpleMDE and `<textarea>`. SimpleMDE could be the default so that existing users wouldn't even notice the change unless they mess with their configuration.

2. Make this part of the UI, so that users can click some sort of button/icon to toggle between SimpleMDE and `<textarea>`.

My preference would be `1.`. What do you think?


cs:
Work in progress. So far, I managed to expose to new config option, so that it is always available in the `MarkdownEditor` component. The code is slightly repetitive. 

@ashmaroli I'm not a React guy, so please don't hesitate to point out if I'm doing something wrong.

ashmaroli:
Woah! this was an issue ticket hours ago.. how did it transform into a pull request?!?! :open_mouth:

ashmaroli:
Couple of tips on the work so far (mainly to be DRY, reduce the diff and ease review..)
- Please revert the change to `package.json` and `yarn.lock`. You should be able to continue work with existing modules (even if you're on Node v12 and above) as long as you don't run `yarn` or `npm install`.
- You may move the repeating code into a new helper function in `src/utils/helpers.js` and import that helper into the view-components.

cs:
@ashmaroli, I integrated your suggestions. Maybe it makes more sense to change the new configuration option to be a boolean? I don't see the need to plan for a third editor option?

Hopefully, I get to the meat of this later today and build the switch between SimpleMDE and `react-textarea-autosize`.

ashmaroli:
> I don't see the need to plan for a third editor option

At the moment I don't see the need either..

Regarding the commit you just pushed (ref: logic in sidebar), I'd prefer that you do not include changes outside the scope of this PR.

cs:
> > I don't see the need to plan for a third editor option
> 
> At the moment I don't see the need either..
> 
> Regarding the commit you just pushed (ref: logic in sidebar), I'd prefer that you do not include changes outside the scope of this PR.

Too many good intentions... fair enough, I remove these changes again. Whatever it takes to get Grammarly support is fine with me ;)

cs:
I will test this a bit to make sure everything works as expected, but at least for the Grammarly use-case, this should be enough. This has been easier than expected ...

ashmaroli:
Let me know once you've finished working on this completely. Since I don't intend to install the grammarly plugin on my browser, I have no sane way to confirm the final render. You could leave a screenshot to give an idea to users that visit this thread in the future.

Regarding styling, I think just using the `field` classname for the textarea should give you the base styles used in the metadata fields. Then your custom styles can be adjusted accordingly..

PR #582, opened 2020-03-05
Prevent unexpected file overwrites for collections, drafts and pages

### LINKED ISSUE #489: Creating a new post will overwrite an existing post when the file already exists
### Description:

Creating a new post will overwrite an existing post sometimes.

This happens when the filename where the post is to be saved is the same as an existing filename.

* Version of JekyllAdmin I'm using <HINT: use `bundle show` to check>: jekyll-admin (0.8.1)
* Version of Jekyll I'm using <HINT: use `bundle show` to check>: jekyll (3.8.5)
* Version of NodeJS I'm using <HINT: use `node -v` to check>: none
* Operating System <e.g. OS X, Windows>: linux
* Browser <e.g, Safari, Chrome>: Chrome

### Steps to reproduce:

1. Create a new (first) post with the title as "test" and no path.
2. Add some content to the first post.
3. Create the first post.
4. Then, create another new (second) post also with the title "test" and no path. This should be done close enough in time to the first one, i.e: the same day usually.
5. Add some different content to the second post.
6. Create the second post.

### I expected the following:

The second post should not not overwrite the first post.

The first post will be replaced by the second post, and the first post is lost.

### Other details:

This isn't quite so bad with posts as the filename usually contains a date, so it's less likely to accidentally happen. 

With collections though which the title can be the only element that ends up in the filename, it's easier to overwrite an existing item in a collection. (repeat steps above with a collection other than posts that doesn't have a date in the file name / path.)

Overwriting the file makes sense for editing existing posts, but I don't think it makes sense when creating new posts. (Same for collections.)

alzeih:
For collections, drafts and pages, return a 404 if a new or renamed item will overwrite an existing item.

Fixes #489 .

Let me know if this looks like the right approach.

Two existing tests were altered to match what I was seeing during manual testing. New tests not written yet, I'll wait until I get some feedback on the changes.

### DIFF
--- a/lib/jekyll-admin/path_helper.rb
+++ b/lib/jekyll-admin/path_helper.rb
@@ -53,9 +53,7 @@ module JekyllAdmin
 
     # Is this request creating a new file?
     def new?
-      return false if request_payload["path"]
-
-      ensure_leading_slash(request_payload["path"]) != relative_path
+      !request_payload["path"]
     end
 
     private
--- a/lib/jekyll-admin/path_helper.rb
+++ b/lib/jekyll-admin/path_helper.rb
@@ -51,6 +51,13 @@ module JekyllAdmin
       ensure_leading_slash(request_payload["path"]) != relative_path
     end
 
+    # Is this request creating a new file?
+    def new?
+      return false if request_payload["path"]
+
+      ensure_leading_slash(request_payload["path"]) != relative_path
+    end
+
     private
 
     # Returns the path to the requested file's containing directory
--- a/lib/jekyll-admin/file_helper.rb
+++ b/lib/jekyll-admin/file_helper.rb
@@ -77,7 +77,11 @@ module JekyllAdmin
     end
 
     def ensure_not_file(file)
-      render_404 unless file.nil?
+      return if file.nil?
+
+      Jekyll.logger.warn "Jekyll Admin:", "Could not create file."
+      Jekyll.logger.warn "", "Path #{file.relative_path.inspect} already exists!"
+      render_404
     end
 
     def ensure_directory
--- a/lib/jekyll-admin/file_helper.rb
+++ b/lib/jekyll-admin/file_helper.rb
@@ -52,7 +52,7 @@ module JekyllAdmin
       ensure_file(written_file)
     end
 
-    def ensure_not_written_file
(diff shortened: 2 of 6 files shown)


lib/jekyll-admin/path_helper.rb
```diff
@@ -51,6 +51,13 @@ def renamed?
       ensure_leading_slash(request_payload["path"]) != relative_path
     end
 
+    # Is this request creating a new file?
+    def new?
+      return false if request_payload["path"]
+
+      ensure_leading_slash(request_payload["path"]) != relative_path
+    end
+
```
ashmaroli: I'm a bit *unsure* about the duplicated logic in `:new?` and `:renamed?`..

Both methods ultimately test `ensure_leading_slash(request_payload["path"]) != relative_path` explicitly. But in the server namespaces, you have `ensure_not_written_file if new? || renamed?`.
So in `:new?` we test if `request_payload["path"]` is a truthy, and if so immediately call `:renamed?` method where `request_payload["path"]` is again tested for a truthy and then proceed if so..

Needs a bit of a DRYing up and refactoring, IMO.

ashmaroli:
@alzeih If you wish to have this shipped in the upcoming release, please find time to make the requested changes a.s.a.p.
Thank you.

lib/jekyll-admin/path_helper.rb
alzeih: I have replaced

```
def new?
  return false if request_payload["path"]

  ensure_leading_slash(request_payload["path"]) != relative_path
end
```
```
def new?
  !request_payload["path"]
end
```

As `new?` doesn't need to compare the value, it only checks that it exists. So this becomes DRY for free.

alzeih:
> * Needs a bit of a refactoring..

Done.

> * Are resources at the `/data/` namespace immune to the issue faced by other types?

No, they are not, but the front end requests for data are inconsistent with those of collections, drafts, and pages, as it doesn't send the path. I can see how to fix it, but it's a separate PR.

ashmaroli:
@alzeih I'm on board with the current approach. However, I think the intent would be better expressed if the `public` methods were named along the lines of `:ensure_not_existing_file` instead.

If you can add more tests, the better.

PR #590, opened 2020-03-19
Fix listing static files in index view

mertkahyaoglu:
Since directories do not have `extname`, when switched to index listing, `FilePreview` was throwing an error saying `file.extname` is undefined. Fixed it by checking file type before accessing `extname`.

### DIFF
--- a/src/components/FilePreview.js
+++ b/src/components/FilePreview.js
@@ -34,11 +34,20 @@ export default class FilePreview extends Component {
     }
   }
 
+  get isImageFile() {
+    const {
+      file: { type, extname },
+    } = this.props;
+    if (type === 'directory') {
+      return false;
+    }
+
+    return /png|jpg|gif|jpeg|svg|ico/i.test(extname.substring(1));
+  }
+
   render() {
     const { onClick, file, splat } = this.props;
-    const extension = file.extname.substring(1);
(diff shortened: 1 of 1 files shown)


src/components/FilePreview.js
```diff
@@ -34,11 +34,20 @@ export default class FilePreview extends Component {
     }
   }
 
+  get isImageFile() {
```
ashmaroli: Why not `isImageFile = () => {}` in consistent with the style used elsewhere in the front-end classes here..?

mertkahyaoglu: It's not a utility method actually, it's local to this class. As the component name implies this component will preview a file but is that file really an image. If this component were to accept more than one files, it would be better to make it a utility.
In addition, I love using these and didn't want to write a function call in render method.

ashmaroli: I mean why the need for `get`..? We don't use `static` for static methods..

mertkahyaoglu: Because it's a getter? This is not a static method. It's a class attribute. Didn't get the problem here. Why do you think we should use `get` except compatibility with the rest of the code? I didn't use it back then because it was not supported that time and now it is.

ashmaroli: > except compatibility with the rest of the code

Just this. I thought explicit `get` was unnecessary. Maybe I'm wrong..

PR #592, opened 2020-03-29
Release v0.10.1

mertkahyaoglu:


ashmaroli:
Um.. Looks like you need to update the version in the test snapshots too..

PR #597, opened 2020-05-05
Allow default metadata fields to be overridden and edited in place

### LINKED ISSUE #596: Is default, editable metadata on a new document possible?
Hi, basic question here: is it not possible to specify an array of "default" editable metadata/frontmatter fields when creating a new document? For instance, let's say I have a collection entitled `faculty_members`:

`_config.yml`
```yaml
collections:
  faculty_members:
    output: true
```

and every `faculty_member`, has the same front matter, like so:

`/_faculty_members/FooBar.md`
```yaml
---
name: Foo Bar
department: Baz
---
Foo Bar is the head of Baz.
```
when I click New Document in `Faculty_members`...
Screen Shot 2020-04-29 at 8 37 02 PM

...is there really no way to have `department` automatically added as a metadata field at the bottom of the new document form? I attempted a few different things, including adding default values to the `_config.yml`, but these fields became read-only and I couldn't edit them, making them useless for this.

I see a few other issues that seem to mention this idea, I _think_, but they all seem to have been closed or ignored, like https://github.com/jekyll/jekyll-admin/issues/464. Am I missing something vital? This seems like a very, very basic ask and is it really impossible?

Thanks for your time.

i-a-n:
Resolves #596 

override-revert

First PR, happy to receive any additional pushes, and sorry for any inconsistencies.

This allows a user to see the default fields for a new or existing document, choose to "override" the defaults, and revert back to the default at any time. The way I accomplished this in general was by removing the `StaticMetaFields` component, and instead allowing `MetaFields` to display both default and additional fields. By checking the redux store (specifically the `metadata: { metadata:{} }` object), the `MetaFields` component will render each field as either a `Static` or regular `MetaField`, and by passing in props, each subcomponent is able to tell whether it is a default field, and whether it is currently in "default mode".

Choosing to "override" a default simply adds that key/value to the metadata redux object, allowing it to be rendered as a regular `MetaField`. "Reverting" it simply removes it, and it goes back to being rendered as a `StaticMetaField`.

Note: I updated snapshot testing and added action/reducer tests, but added no unit tests. I'd appreciate any help or direction on that, if you feel we need them. Thanks again.


ashmaroli:
Good job, @i-a-n! The code looks good to me on first glance..

src/components/metadata/StaticMetaButton.js
```diff
@@ -0,0 +1,48 @@
...
+export default class StaticMetaButton extends Component {
+  state = {
+    dropdown: false,
+  };
+
+  toggleDropdownState = () => {
+    this.setState(state => {
+      return { dropdown: !state.dropdown };
+    });
+  };
```
mertkahyaoglu, line 15:
```suggestion
  toggleDropdownState = () => {
    this.setState(state => ({ dropdown: !state.dropdown });
  };
```

src/components/metadata/MetaButtons.js
```diff
@@ -36,23 +36,30 @@ export default class MetaButtons extends Component {
...
+      .filter(Boolean);
   }
 
   render() {
-    const { currentType, parentType, onRemoveClick } = this.props;
+    const {
+      currentType,
+      parentType,
+      onRemoveClick,
+      isADefaultField,
```
mertkahyaoglu, line 61: Let's rename this to `isDefaultField`.

mertkahyaoglu:
@i-a-n We can merge this after tests are passing.

src/components/metadata/MetaButtons.js
```diff
@@ -94,4 +101,5 @@ MetaButtons.propTypes = {
   onConvertClick: PropTypes.func.isRequired,
   onRemoveClick: PropTypes.func.isRequired,
   parentKey: PropTypes.string,
+  isDefaultField: PropTypes.bool.isRequired,
```
ashmaroli, line 104: Why is this a `required` attribute?
IMO, if it was optional, then you needn't have to explicitly `isDefaultField={false}` at so many places.
And if it should be `required`, then please move it above one line so that required parameters are together..

i-a-n: good idea, I don't see a reason why it needs to be required. let me update.

ashmaroli:
Thank you @i-a-n

PR #617, opened 2020-12-04
Add option to change homepage

### LINKED ISSUE #156: Dashboard main page, rather than being welcomed by posts or pages

mertkahyaoglu:
Closes #156

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,9 +28,7 @@ Refer to the [installing plugins](https://jekyllrb.com/docs/plugins/installation
 
 ## Options
 
-Jekyll Admin related options can be specified in `_config.yml`
-under a key called `jekyll_admin`. Currently it has only one option `hidden_links`
-which is for hiding unwanted links on the sidebar. The following keys under `hidden_links` can be used in order to hide default links;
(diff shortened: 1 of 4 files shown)


ashmaroli:
What if the user would like to have the `data` or `drafts` route as their homepage?

mertkahyaoglu:
@ashmaroli I didn't think anybody would want that 😅 We can add that but do you think that's a useful case?

ashmaroli:
Since users are asking for the homepage to be something other the `/pages/` route, I think it'd be best if user could choose whatever supported route to be their indexRoute..

mertkahyaoglu:
@ashmaroli 👍 Added all other routes as well.

ashmaroli:
The implementation LGTM!
Does this need test coverage?

mertkahyaoglu:
Absolutely but it's a bit tricky to test so I skipped it thinking code is safe. We'll see if we get any issue about this 😅

ashmaroli (approved):

Issue #630, opened 2021-05-02
Broken tests against Jekyll 4.2


coreycarvalho:
### Description:
Jekyll 4.2 introduced attribute caching  here https://github.com/jekyll/jekyll/pull/8497/files, which causes content to be returned differently. I'm not sure if this was intended to return differently when cached, but it's there nonetheless and we should modify our tests to accommodate. As it stands, this doesn't allow development using Jekyll 4.2.

### Tell us a bit about yourself:
* Version of JekyllAdmin I'm using: 0.11.0
* Version of Jekyll I'm using: 4.2
* Version of NodeJS I'm using: 12.22.1
* Operating System: Fedora 33 on WSL

### Steps to reproduce:
* Force Jekyll 4.2 to be installed in Gemfile
* Run `script/cibuild-ruby`
* Watch content related tests fail

### I expected the following:
Expected all tests to pass from master

### But got the following, instead:
```
  1) JekyllAdmin::APIable page with content includes the rendered content
     Failure/Error: expect(content).to eql(expected)

       expected: "<h1 id=\"test-page\">Test Page</h1>\n"
            got: "# Test Page\n"

       (compared using eql?)

       Diff:
       @@ -1 +1 @@
       -<h1 id="test-page">Test Page</h1>
       +# Test Page

     # ./spec/jekyll-admin/apiable_spec.rb:33:in `block (6 levels) in <top (required)>'

  2) pages getting a single page returns the rendered output
     Failure/Error: expect(last_response_parsed["content"]).to eq(expected)

       expected: "<h1 id=\"test-page\">Test Page</h1>\n"
            got: "# Test Page\n"

       (compared using ==)

       Diff:
       @@ -1 +1 @@
       -<h1 id="test-page">Test Page</h1>
       +# Test Page

     # ./spec/jekyll-admin/server/page_spec.rb:99:in `block (3 levels) in <top (required)>'

Finished in 17.59 seconds (files took 0.59848 seconds to load)
205 examples, 2 failures
```

coreycarvalho:
I attempted to address this in my PR. IMO, this seems pretty important to allow new PRs to pass tests.

coreycarvalho:
The PR to address this issue has been open for a little while - any chance I can get some eyes on this?

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

[...]

PR #631, opened 2021-05-02
Fixes broken tests from 4.2

coreycarvalho:
This PR addresses https://github.com/jekyll/jekyll-admin/issues/630

The issue linked describes the issue.

For the `apiable` test, I opted to change the expected value to what liquid returns. This accounts for potential changes that Jekyll makes to the exact content returned.

For the `page` test, I opted to match against "Test Page", since we can reasonably expect that to be returned regardless of if Jekyll changes the exact values returned.

### DIFF
--- a/spec/jekyll-admin/apiable_spec.rb
+++ b/spec/jekyll-admin/apiable_spec.rb
@@ -14,6 +14,8 @@ describe JekyllAdmin::APIable do
 
       [false, true].each do |with_content|
         context "#{with_content ? "with" : "without"} content" do
+          let(:as_liquid) { subject.to_liquid }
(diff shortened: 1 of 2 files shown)


coreycarvalho:
Anyone around to review this? 😬

ashmaroli:
@coreycarvalho Thank you for your contribution.
Unfortunately, I think this change is no longer necessary.

Could you try reproducing #630 with `jekyll-4.2.1` at your end?

coreycarvalho:
@ashmaroli hey - it looks like this issue is indeed fixed with `jekyll-4.2.1`

PR #639, opened 2021-05-25
Translated constants to Chinese

albertix:
Translated current English language constants to Chinese


mertkahyaoglu:
@albertix Thanks for the pr but i18n is not active right now. We can keep this for future.

PR #684, opened 2022-05-16
Add option to set admin specific scheme/host/port

### LINKED ISSUE #610: Jekyll url is not applied in View button
# Description:

And pressing the `View` button in the admin panel the URL goes to http://localhost:4000. But I'm behind a reverse proxy called Nginx.

* Version of JekyllAdmin I'm using:  0.10.2
* Version of Jekyll I'm using: 4.1.1
* Version of NodeJS I'm using: v14.15.0
* Operating System <e.g. OS X, Windows>: Linux!
* Browser <e.g, Safari, Chrome>: Firefox!

### Steps to reproduce:

1. Setup Jekyll as normal
2. Configure `_config.yml` file, and set `url` key  to: `url: "https://yourdomain.com"` (`baseurl` is just "/")
2. Configure Nginx to use `yourdomain.com` as server_name in Nginx with a `location /` section containing at least: `proxy_pass http://localhost:4000;`
3. Go to admin panel: https://yourdomain.com/admin
4. Press the View button
5. Links goes to `http://localhost:4000/jekyll/update/2020/11/19/welcome-to-jekyll.html`, which will **NOT** work.

### I expected the following:

I expect the View links to take the url settings into account. The link example above, should go to:

https://jekylltest.harmkeslaatdoor.nl/jekyll/update/2020/11/19/welcome-to-jekyll.html

http://localhost:4000/jekyll/update/2020/11/19/welcome-to-jekyll.html

I think it shouldn't be that hard to get this solved, I hope 👍🏽 ?

Regards,
Melroy van den Berg

CGillen:
Fixes #610 
@ashmaroli mentioned they wanted tests on the fix. These might not be the best tests, so if you have any suggestions I'll gladly fix them up.

### DIFF
--- a/docs/configs.md
+++ b/docs/configs.md
@@ -1,7 +1,7 @@
 ---
 title: Configuration Options
 permalink: /configs/
---- 
+---
 
 Jekyll Admin related options can be specified in `_config.yml`
 under a key called `jekyll_admin`.
@@ -10,7 +10,7 @@ under a key called `jekyll_admin`.
 
 #### `hidden_links`
 
-For hiding unwanted links on the sidebar. 
+For hiding unwanted links on the sidebar.
 
 The following keys under `hidden_links` can be used in order to hide default links:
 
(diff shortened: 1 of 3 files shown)


ashmaroli:
@CGillen Thank you for proposing an enhancement.

From a quick glance, it looks like you're assuming `site.config["jekyll-admin"]` will always be defined.
What if it isn't?

CGillen:
`#dig` would be a better solution, wouldn't it? Doubtful `site.config["jekyll_admin"]` would be defined as anything besides a hash.

PR #692, opened 2022-10-31
Added links buttons in header

rpeyron:
Hello,

I am a new user to jekyll and I love jekyll-admin

This PR adds the ability to configure in _config.yml custom buttons in header to have some useful links when editing your site (for instance, I will add links to some free image libraries, to my web analytics, etc.)

image

#### `header_buttons`

Add buttons in the header to open custom links. You may provide for each link:
- `title`: the button label
- `url`: the URL to be opened
- `icon`: the fontawesome icon to be used (if none, the view icon will be used)

```yaml
jekyll_admin:
  header_buttons:
   - title: Google
     url: 'https://www.google.com'
     icon: search
```

Hope you will find this useful

### DIFF
--- a/docs/configs.md
+++ b/docs/configs.md
@@ -35,3 +35,19 @@ Valid values for `homepage`: `pages` (default), `posts`, `<collection_name>`,
 jekyll_admin:
   homepage: "posts"
 ```
+
+
+#### `header_buttons`
+
+Add buttons in the header to open custom links. You may provide for each link:
+- `title`: the button label
+- `url`: the URL to be opened
+- `icon`: the fontawesome icon to be used (if none, the view icon will be used)
+
+```yaml
+jekyll_admin:
+  header_buttons:
+   - title: Google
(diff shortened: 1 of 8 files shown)


ashmaroli:
Thank you for proposing this enhancement, @rpeyron but unfortunately, I don't see the value in this for the user-base as a collective.
Therefore, I am closing this.
If you can come up with a better use case, I will consider re-opening.

rpeyron:
Ok, I understand you want to keep the codebase as low as possible. Do you plan a plugin system to be able to add user specific configurable features?

ashmaroli:
To clarify, I do not *intend to keep the codebase low*. A new feature *has to* be beneficial for a significant section of the user-base.
The proposed enhancement here probably has value only to you. Other users will have to wonder what they would do with the buttons in the header.

This project was not designed to be extendable via plugins. The only course is to fork, customize code, build and deploy for custom use.

PR #698, opened 2022-12-11
Fix duplicated entries bug

rpeyron:
I was getting duplicated entries on my jekyll-admin instance for each content type (drafts, static files,...). I was not able to debug the ruby code to understand what was going on and it does not occur with the test site of jekyll-admin. But by replacing .entries method of Pathname with [.children](https://ruby-doc.org/stdlib-2.6.3/libdoc/pathname/rdoc/Pathname.html#method-i-children) (that do not include dot directories) I have no more the duplicated entries issue.

### DIFF
--- a/lib/jekyll-admin/directory.rb
+++ b/lib/jekyll-admin/directory.rb
@@ -12,9 +12,8 @@ module JekyllAdmin
     include JekyllAdmin::APIable
 
     RESOURCE_TYPES  = %w(pages data drafts static_files).freeze
-    DOT_DIRECTORIES = [".", ".."].freeze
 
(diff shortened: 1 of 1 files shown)


ashmaroli:
By calling `next, we skip the current iteration. Therefore, the proposed change does not make sense. In other words, this doesn't really fix anything.
*You may still encounter duplicate entries* at random. Keep making random edits to your source files to trigger this flaky elusive bug.

rpeyron:
You are perfectly right that was not making sense, and indeed the duplicates came back. I was not conscious it was kinda random, I will keep investigating and try to learn how to debug ruby code (the only thing I am sure about is that the duplication occurs on the server part). Sorry

PR #700, opened 2022-12-17
Process site only if not in `watch` mode

### LINKED ISSUE #699: Files are displayed duplicated after saving
### Description:

Files are displayed duplicated after saving with jekyll-admin.

Example:
image

* Version of JekyllAdmin I'm using <HINT: use `bundle show` to check>: 0.11.1
* Version of Jekyll I'm using <HINT: use `bundle show` to check>: 4.3.1
* Version of NodeJS I'm using <HINT: use `node -v` to check>: v16.18.1
* Operating System <e.g. OS X, Windows>: Linux Debian 11
* Browser <e.g, Safari, Chrome>: Chromium 108

### Steps to reproduce:

Note: I have not been able to reproduce with the spec/fixtures test site ; maybe it is too fast to generate to have concurrency issues

1. Launch with bundle exec jekyll serve --live-reload
2. Edit a file
3. Save the file

### I expected the following:

Files are not shown duplicated

Files are shown duplicated

The console also reports other errors as:
`conflicting chdir during another chdir block`
or content duplication warning with same source file
```
Conflict: The following destination is shared by multiple files.
                    The written file may end up with unexpected contents.
                    /home/remi/.roaming/work/lprp/rpeyron.github.io/_site/files/triboum.pdf
                     - /home/remi/.roaming/work/lprp/rpeyron.github.io/files/triboum.pdf
                     - /home/remi/.roaming/work/lprp/rpeyron.github.io/files/triboum.pdf
```

and site.collections contain all entries duplicated

### Other details:

Workarounds that I found:
- run with --no-watch :  `bundle exec jekyll serve --live-reload --no-watch`
- or comment the `site.process ` commands in file_helper.rb

I guess the two lines above are triggering the same process that may cause problems when run concurrently

rpeyron:
Fix mentioned in #699 

Will avoid duplicate site processing: either we are in watch mode and jekyll-admin do not need to trigger site.process, or we are not in watch mode and jekyll-admin will trigger site.process

Note that it does not work when using `--livereload --no-watch` options simultaneously, which has little interest, as livereload option will force `config{"watch"]=1`.

Closes #699

### DIFF
--- a/lib/jekyll-admin/file_helper.rb
+++ b/lib/jekyll-admin/file_helper.rb
@@ -22,14 +22,14 @@ module JekyllAdmin
       File.open(path, "wb") do |file|
         file.write(content)
       end
-      site.process
+      do_process
     end
 
     # Delete the file at the given path
     def delete_file(path)
       Jekyll.logger.debug "DELETING:", path
       FileUtils.rm_f sanitized_path(path)
-      site.process
+      do_process
     end
 
     def delete_file_without_process(path)
@@ -38,6 +38,10 @@ module JekyllAdmin
     end
 
     private
+
+    def do_process
+      site.process unless site.config['watch']  # to avoid race condition between jekyll watch and this
+    end
 
     def ensure_requested_file
       ensure_file(requested_file)
--- a/lib/jekyll-admin/file_helper.rb
+++ b/lib/jekyll-admin/file_helper.rb
@@ -22,14 +22,14 @@ module JekyllAdmin
       File.open(path, "wb") do |file|
         file.write(content)
       end
-      do_process
+      conditionally_process_site
     end
 
     # Delete the file at the given path
     def delete_file(path)
       Jekyll.logger.debug "DELETING:", path
       FileUtils.rm_f sanitized_path(path)
-      do_process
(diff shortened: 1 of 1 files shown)


ashmaroli:
Thank you for submitting this pull request, @rpeyron.

Leaving a few points as feedback:
- The name `do_process` isn't fully transparent to the underlying implementation. How about `conditionally_process_site`? That way, a future contributor (including us) will know what to expect immediately on reading the code.
- The Admin interface allows one to edit the configuration file as well. So, if I remember correctly, that may allow changing `site.config["watch"]` (without actually loading Jekyll's Watcher). *I haven't tested this; just a hypothesis.* The solution for this *could be* initializing a new object (e.g. `JekyllAdmin.site_processor`). If this is too confusing for you, let me know. I will make that change in a separate pull request / commit.

rpeyron:
Hello @ashmaroli ; thanks for your feedback
- You are right, I have renamed as suggested
- I am not sure to have understood the solution you have in mind, I will leave it to you

ashmaroli (approved):

ashmaroli:
Thank you @rpeyron

